PR #13, opened 2020-06-07
Haskell Web3 milestone 2

akru:

### DIFF
--- a/deliveries/hs-web3-milestone-2.md
+++ b/deliveries/hs-web3-milestone-2.md
@@ -1,0 +1,13 @@
+# Milestone Delivery :mailbox:
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone.**  
+
+* **PR Link:** https://github.com/w3f/Open-Grants-Program/pull/15
+* **Milestone Number:** 2
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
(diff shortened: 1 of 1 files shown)


Noc2:
Hi @akru. Thanks for the delivery of another milestone and sorry for the delay regarding your last delivery. I will look into both as soon as possible (definitely within the next few days).

akru:
@Noc2 Thank you, David! No problem!

Noc2:
I started with the evaluation, but I think the changing of the license file is causing some issues with the examples, [see](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/hs-web3_2_Noc2.md#general-notes). Could you take a look at it or tell me how I can easily resolve this?

akru:
Ah, yes, thank you @Noc2 for the report, fixed in https://github.com/airalab/hs-web3/commit/ff38be7bd40ddd5315b1ffbe842bd87bad45f232,
I've just move `BSD3` license as default LICENSE file because it covers most of the code, Apache-2.0 license covers code tagged in source code headers license string, for example here: https://github.com/airalab/hs-web3/blob/master/src/Codec/Scale.hs#L6

Noc2:
Sorry. Stack for some reason used some old files even though I checked out the latest version. I’m not used to haskell yet (although stack is actually quite similar to cargo). Btw. I couldn’t find the implementation for tuples.

akru:
@Noc2 yes, stack is pretty close to cargo, and LTS in stack is pretty convenient.

Tuple instances implemented using templates here: https://github.com/airalab/hs-web3/blob/master/src/Codec/Scale/TH.hs#L24
Typles in Haskell are same as production type structures, we can use default generic instance.

Noc2:
Thanks for pointing this out. I’m happy to tell you that the milestone is a pass and will forward the invoice.

PR #15, opened 2020-06-08
Java client lib - Milestone 2 delivery

splix:

### DIFF
--- a/deliveries/java_client_mileston_2.md
+++ b/deliveries/java_client_mileston_2.md
@@ -1,0 +1,15 @@
+# Milestone Delivery :mailbox:
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone.**  
+
+* **PR Link:** https://github.com/w3f/Open-Grants-Program/pull/12 
+* **Milestone Number:** 2
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | HTTP Client | https://github.com/emeraldpay/polkaj | Milestone delivers modules `polkaj-json-types`, `polkaj-api-base`, `polkaj-api-http`, and `polkaj-api-ws` which implements all core functionality to access Polkadot APIs | 
+| 2.  | Example App | https://github.com/emeraldpay/polkaj/tree/master/examples | `rpc` example contains several basic examples how to access Polkadot RPC, make calls, and subscribe to changes. `runtime-explorer` provides a web explorer of Runtime Metadata, that could be helpful for building actual RPC requests | 
+| 3.  | Unit Tests |https://codecov.io/gh/emeraldpay/polkaj (coverage) | Tests are in each module in src/tests/ dir. The current coverage is 89% | 
+| 4.  | Documentation | https://github.com/emeraldpay/polkaj/tree/master/docs | New sections covering RPC, plus minor updates to other sections | 


Noc2:
Thanks again for the delivery. We will try to look into it as soon as possible.

mmagician:
Thanks for the delivery. Could you provide a suite of integration tests as well? 
The way I'd see this:
- instructions on how to run a local node(s) and any special configuration that needs to be passed. Links to substrate tutorials etc. would be fine (as long as any special configuration is also provided by you)
- smoke tests to assert that we can interact with the node & listen to events

I also made a PR with some cosmetic changes, feel free to have a look and add changes:
https://github.com/emeraldpay/polkaj/pull/14

splix:
Thank you @mmagician 

For the node just running `polakdot` without any options should be enough, like:
```
$ polkadot
2020-06-12 11:27:30 ----------------------------
2020-06-12 11:27:30 This chain is not in any way
2020-06-12 11:27:30       endorsed by the
2020-06-12 11:27:30      KUSAMA FOUNDATION
2020-06-12 11:27:30 ----------------------------
2020-06-12 11:27:30 Parity Polkadot
2020-06-12 11:27:30 ✌️  version 0.7.34-dev-27b201ac-x86_64-macos
2020-06-12 11:27:30 ❤️  by Parity Technologies <admin@parity.io>, 2017-2020
```

As a smoke test with a real node, instead of http mocks used in unit tests, you can run examples from `examples/rpc` and `examples/runtime-explorer`. There is instruction in `examples` dir

splix:
@mmagician maybe I misunderstand, and you actually insist in making an integration testing suite? I thought it's a question, but I'm not sure now

mmagician:
> @mmagician maybe I misunderstand, and you actually insist in making an integration testing suite? I thought it's a question, but I'm not sure now

In general I'd love to see an integration suite, but I will not insist on integration tests in this case. It would be unfair given that you clearly stated in milestone requirements that you'd be including examples & they seem to indeed illustrate the functionality well. That said, I still have a couple of action points:

- could you please make the instructions about running the node a bit more explicit (perhaps with an example)? I imagine that the current: `The examples expect Polkadot node running and listening for RPC and WS requests on the localhost (i.e. default node options)` might be somewhat unclear to a complete newcomer. Or perhaps include links to relevant substrate pages etc. that would allow a user to setup a node.
- `gradle run -PmainClass=<FollowState/DescribeRuntime>` fails for me with: `Error: Could not find or load main class extension 'application' property 'mainClass'`. I need to apply the solution from: https://stackoverflow.com/a/21358586, i.e. remove the outer application tag. Could you advise me on whether I am missing something or else please update the example.
- FollowState example fails to run. It seems that the first message of the form :`{"jsonrpc":"2.0","result":"0YrW7zAp11yfkgPu","id":0}` cannot be deserialised. I'd be happy to provide more logs in case you can't reproduce this locally.

splix:
Technically you don't need to have any special config. Just running the binary `polkadot` as is, is enough for the examples. That's the whole instruction, I guess. You don't need to do anything special, no options, no configs, nothing.

But I see your point that the user may be unfamiliar with installing and running the Polkadot. It's not described in the library docs right now. I was writing the docs with the assumption that the library will be used by people who are already know how to run the Polkadot. It's a good idea to describe it how to download and install Polkadot, thank you for the suggestion.

For the Gradle issue, I suspect that you have a different version of Gradle. So now I have packaged the Gradle wrapper within the examples and updated instructions that will use it. That will force to run examples with a specified version of Gradle, and I hope that will fix the issue.

Regarding the String `"WUz5T1PxFoLtfXIt": not a valid Integer value`. Apparently, the Polkadot API has changed subscription ids from numbers to string at some point. I was using a previous version for the development and wasn't aware of it. It's now fixed, and the error is gone now with the new version of the lib.

splix:
@mmagician hi, did it resolve the issues you had?

mmagician:
@splix I looked at your new commits quickly, it seems like it should fix most of the points I mentioned 👍 

I will do a throughout update of my evaluation before the end of the week.

mmagician:
@splix I've added an evaluation.
https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/java-client_2_mmagician.md

Please have a look and address the parts about `Completeness` and `Documentation`.

splix:
Hi @mmagician. Thank you for the evaluation.

- RPC coverage. It in fact doesn't cover all of the possible RPC methods. But technically all of the methods are available, the library just doesn't give a nice Java wrapper, and only because those wrappers would need some of the cryptography left for Milestone 3. So you can use all of the RPC methods already, if you have some 3rd party or own code to do crypto, for example.

- I'm going to extend documentation, and add integration test. But just for your understanding, there is just little problem with setting up everything for integration tests, since it should be fully automated and consistent, also raises questions how much the blockchain must be synchronized for integration testing, etc. I mean it's possible, and I'm hoping I'll add that, but it's not an easy fix, it's something that may take a week or so.

mmagician:
@splix As mentioned, I will **not** require the integration tests for this delivery, I just mentioned them as nice-to-haves especially in the context of that spec change in Polkadot API. If you want to develop them as extra, I can only encourage you :)

However, to approve this delivery I would like to see the full coverage of RPC methods that you specified in the original proposal, namely:
- chain
- contracts
- state

These should match the corresponding sections in https://polkadot.js.org/api/substrate/rpc.html

splix:
Thank you @mmagician. I understand your concerns and will cover the other RPC methods with temporary wrappers. 

To clarify, even right now, you can execute any of the RPC methods, so technically, you can't say some of the methods are not covered. It seems that the main issue is that some of them don't have a special name at this moment, and, for example, hashing is not integrated, which is going to be part of Milestone 3. Without it, the methods will be useless but can be implemented to comply with the requirements.

splix:
@mmagician please see the latest version

mmagician:
@splix Thank you for the changes, I will review when possible.

splix:
@mmagician hi. Did you have a chance to take a look?

mmagician:
@splix not yet, I will keep you posted

mmagician:
@splix Thanks very much for your work to create the wrappers. I understand your argument about users being able to use the RPC calls directly, although I would argue that if we're creating wrappers in the first place (which, as you said, are not **needed** per se) then we might as well create a full suite of them for the convenience. If I were to write an application on top of your library, I'd rather not mix `PolkadotApi.commands().<some_method>` with `<RpcCall.create(<some_method>)` but be consistent with my usage.

That said, I encountered just one small issue that perhaps you could help me resolve. In `ShowState.java` I tried:

```
client.execute(PolkadotApi.commands().getFinalizedHead());
client.execute(PolkadotApi.commands().getBlockHash());
```
Now `getFinalizedHead` works just fine. 
However I get an error when trying to call the `getBlockHash` wrapper:
```
> Task :compileJava FAILED
/Users/marcin/web3/polkaj/examples/rpc/src/main/java/ShowState.java:43: error: cannot find symbol
        client.execute(PolkadotApi.commands().getBlockHash());
                                             ^
  symbol:   method getBlockHash()
  location: class StandardCommands
1 error

FAILURE: Build failed with an exception.
```

Maybe it's just my setup? I tried clearing out gradle cache & build directories but that didn't help. I couldn't quite grasp why it would not find the method. 

Other than this it's looking good.

Thanks for help!

splix:
The main reason why not everything is shortcut is the difference between Java and Javascript. Later is a procedural language with passing unstructured entities, which gives simplicity and JSON. But Java is an Object-Oriented programming language, and designing a library in a procedural style, based on `static` methods, is not the very right way. 

Most of the `state_` RPC is a good example of why. It's the methods that depend on a context, like "which state", and supposed to be called from such context-aware object. I.e., it's most likely that the application will have part of the parameters, such as `key`, hid as `private` inside an object, and expect only additional part of the query. Also, it's very unlikely that the developer will want to operate data on a byte level because in Java you usually operate objects. This means that in most cases, you'll build a `RpcCall` with transformers depending on your particular task. It may make more sense to have a Builder for that, though it is not clear yet how exactly it supposed to be used, but a static shortcut is more like a distraction.

Shortcuts in StandardCommands are useful when you have a simple non-extendable call that you may use in many parts of the code and without any context. Like `getBlockHash()` is the best candidate, you don't need anything else, so a static shortcut saves some keystrokes and makes sense.

Please also note that shortcuts, which is the main topic of our arguing in the past weeks, and a reason for rejecting the milestone, is not really a part of deliverables. It is an addition. Since, formally, the deliverables list includes _HTTP Client_ only, which is covered by `PolkadotApi`, `PolkadotSubscriptionApi`, `RpcCall`, `Subscription`, and their implementations. As they cover all possible ways of consuming API. I appreciate your feedback, and also want to make the best library for Polkadot, that's why I added the shortcuts following your advice, even though I disagree with that design, but still find it sad that the main arguing is not about the core of the milestone, but rather about diffrent views on designing additional stuff.


Regarding the examples. It's strange that `./gradlew run -PmainClass=ShowState` doesn't work, since the "missing" method (from the error message) is clearly in place, and was here untouched for a long time. Maybe the library got corrupted somehow? If you didn't edit the sources, then may be a compiled library got corrupted? It's highly unlikely, but I don't see any other reason. You may delete all compiled binaries by `rm -f ~/.m2/repository/io/emeraldpay/polkaj` and try again.

mmagician:
@splix That's a nice explanation, I appreciate it. 
I've tried removing the binaries but that didn't help. I also tried cloning the repo again and trying from scratch - still no luck.
What I'm trying to accomplish is in this branch: https://github.com/mmagician/polkaj/tree/get-block-hash

Could you try it out yourself and see if you manage to run `./gradlew run -PmainClass=ShowState` without problems?

splix:
I've checked it, and it turns out it was my bad. I confused the method with another one, also forgot that the latest changes weren't published yet. Please see the latest version of the project, I've added your line, and it works now.

splix:
I think I found the missing part. The RPC methods should be documented, regardless if they have a shortcut or not. So I've added a go-to class for method reference and documentation. That would be helpful as being a part of the Javadoc, and more importantly, will be shown in IDE when you call a method. See https://github.com/emeraldpay/polkaj/commit/c58c9c88942a2274db781e66b77f8de35f25ed25

splix:
@mmagician hi, did you have a chance to update the evaluation? It seems we both agree now that the current evaluation is based on a misunderstanding of the code, so it would be great if you find a time to fix it.

mmagician:
@splix I will be looking into it soon & will let you know once I have an update.

splix:
Thank you. Though I would also appreciate it if you update the current evaluation, since, I'm sure, you understand now, that it contains an incorrect statement about missing HTTP Client. I mean, I understand you want to spend more time learning the code and so on, but as we already figured out the misunderstanding about the HTTP Client, it may be removed, regardless of the other things.

Noc2:
Hi @splix Thanks for fixing the issues and your help with the evaluation. Marcin just updated his [notes](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/java-client_2_mmagician.md) and accepted your delivery. I will forward your invoice internally.

splix:
@Noc2 That's the great news. I'm starting with the Milestone 3 then. Thank you

RouvenP:
hi @splix we just sent a test transaction. could you please confirm if received?

splix:
@RouvenP yes, I confirm receiving of 0.0001 BTC, thank you

RouvenP:
@splix thanks for confirming, transfer is done.

splix:
@RouvenP I confirm receiving, thank you!

PR #24, opened 2020-07-13
dipole-oracle-millestone-1

renyanduo:

### DIFF
--- a/deliveries/dipole-oracle-millestone-1.md
+++ b/deliveries/dipole-oracle-millestone-1.md
@@ -1,0 +1,17 @@
+# Milestone Delivery :mailbox:
+
+> Only the GitHub account, which is responsible for the pull request of the accepted application is allowed to submit milestones. Don't remove any of the mandatory parts presented in bold letters or as headlines!
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone.**  
+
+* **PR Link:**  [DipoleOracle](https://github.com/w3f/Open-Grants-Program/pull/31). 
+* **Milestone Number:**  [Milestone1](https://github.com/w3f/Open-Grants-Program/blob/master/applications/DipoleOracle.md#Milestone)
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | DipoleOracle |https://github.com/DipoleTech/dipole-oracle|  |
+| 2.  | Backend Client |https://github.com/DipoleTech/dipole-oracle/tree/master/examples/nodejs|  |
+| 3. | Documentation |https://github.com/DipoleTech/dipole-oracle/blob/master/README.md| |
+


Noc2:

mmagician:
@renyanduo I've started looking at your delivery.

Before doing a deep dive into the code, I would like to point out a couple of (equally important in open source projects) issues about the delivery in general.

1. Regarding documentation, in [this](https://github.com/w3f/Open-Grants-Program/blob/master/applications/DipoleOracle.md) document you started describing the roles of Collector, Operator etc. This information is completely missing from your repo - I think it would be a valuable addition to your README, especially if it could be expanded further.

2. As a person without extensive knowledge into the energy industry, I've had a hard time time understanding the nodejs example provided. I would like to see a more extensive instruction set there, including:
- What is the goal of the example? What is the expected output and how should I interpret it?
- What is the setup that I'm supposed to be running? (I just tried with one node as per your README: `./target/release/dipole-oracle --dev`)
- How does each of the four roles: Operator, Collector, PayOracle and GoodsOracle tie into the example? 

3. Comparing your initial application and the delivery, the part `complete benchmarking, weight annotation, test coverage and code linting` is missing altogether, as far as I can tell - or am I overlooking something? 

4. Finally, it is hard to evaluate such an involved project without a real-life example. Would it be possible to get hold of the hardware you're using, so that we could test it in a real use case? That would be the ideal option - secondary one would be for us to have a video call where you to give us a demonstration of the hardware & software.

As I said, these are rather superficial comments from looking at your delivery, but they would immensely aid our understanding of your delivery, as well as help kickstart the potential users' journey with dipole.

renyanduo:
@mmagician   thanks for your suggestion, We have spent a lot of time and fixed the problems above . We updated an easy-understanding example, please have a look and let us know what you think. Many thanks.

mmagician:
@renyanduo Thank you for the updates. I have been able to compile the code and your examples.

I still have some questions:

Operator
- Is my understanding correct, that an operator's main responsibility is to place the offline (off-chain) data from devices onto the blockchain?
- But how do you register a device? Who is responsible for this, a central authority or the users themselves?
- What mechanism is there in place to prevent malicious users from uploading fake data?

Collector
- If I understand well, the Collector gathers the data uploaded by operator, and does some post-processing with it. 
- What is the real advantage of having a Collector vs. just grabbing the data from the blockchain and performing calculations yourself? This way, only the party interested needs to perform calculations, thus saving compute resources.
- Is the raw data or post-processed data stored on the chain?
- Is it possible for you to estimate the amount of data (both raw and processed) coming from a single device (one of the devices that you have envisioned)? I am after daily/hourly volume, update frequency and whatever else you could provide.

Where do GoodsOracle and PayOracle fit into this picture? I saw that you have updated the descriptions for each role, but as mentioned before, these are still vague...Could you expand this further, perhaps with a list of actions that such a role can assume (e.g. register a device). I believe that such a list should be easy to compile, as it corresponds 1:1 (from my understanding) with the pallet implementation. 

Finally, I would suggest that we arrange for a call where you walk us through the setup and answer any outstanding questions. I am also curious to see a demo with real hardware devices. 

You can reach me on Element to coordinate further: @marcin:web3.foundation

renyanduo:
Thanks a lot, we have updated the doc according to your new questions. And we hope to have a video call to explain what you're still  confused, show how  real hardware devices work with our project, and communicate further about the project. We'll contact you on Element.

renyanduo:
@mmagician Hi, we have updated the doc & uploaded a short vedio on youtube, https://youtu.be/Gd3RcsnIN4k.Please have a look and let's know what you think.Many thanks.

PR #32, opened 2020-08-17
Java client lib - Milestone 3 delivery

splix:

### DIFF
--- a/deliveries/java_client_mileston_3.md
+++ b/deliveries/java_client_mileston_3.md
@@ -1,0 +1,22 @@
+# Milestone Delivery :mailbox:
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone.**  
+
+* **PR Link:** https://github.com/w3f/Open-Grants-Program/pull/12 
+* **Milestone Number:** 3
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | schnorrkel/sr25519 module | https://github.com/emeraldpay/polkaj | module `polkaj-schnorrkel` |  
+| 2. | Signing and signature validation module | https://github.com/emeraldpay/polkaj | module `polkaj-tx` |
+| 3. | Example App | https://github.com/emeraldpay/polkaj/tree/master/examples | `keys` example shows how to use Schnorrkel for Secret/Public Keys. `balance` gives exampels on listening to storage changes, getting information about account and making a transfer |
+| 4. | Unit Tests | https://codecov.io/gh/emeraldpay/polkaj (coverage) | Tests are in each module in src/tests/ dir. The current coverage is 86% |
+| 5. | Documentation | https://github.com/emeraldpay/polkaj/tree/master/docs | New sections covering Schnorrkel, Extrinsics and Balance  
+
+With this delivery the Polkaj library covers most of use cases related to access Storage, operate Secret Keys, and create and sign Extrinsics.
+It provides a ready to use wrappers for Balance related operations (get current and make a transfer).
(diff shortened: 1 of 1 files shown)


mmagician:
@splix Thanks for the next milestone, we'll look at it in the upcoming days!

mmagician:
@splix The delivery looks very good so far. I am still working my way through the code, but just two small things I found till now:
* RPC examples are broken now (I checked against v0.2.3 where it run fine)
* gradle.wrapper is missing from `keys` examples (after I've added it manually it run fine)

I will let you know if I have more questions. Thanks!

splix:
@mmagician oh, I didn't check the old examples, will update them to make sure they work with the latest version. Thank you for bringing it in. And will put the wrapper as well.

mmagician:
@splix Out of curiosity, why did you decide to encode the Era byte array in the reverse [order](https://github.com/emeraldpay/polkaj/blob/master/polkaj-tx/src/main/java/io/emeraldpay/polkaj/tx/Era.java#L106) to the [reference](https://github.com/polkadot-js/api/blob/f43bc32573/packages/types/src/extrinsic/ExtrinsicEra.ts#L176) implementation? (I'm basing my assumption that this was your reference, on the doc comments which are the same as in polkadot-js repo). 

Would it not be less confusing (and less error prone for users?) to keep the same format? I find the Era encoding/decoding challenging enough as it is :) 

Let me know your thoughts!

mmagician:
@splix Another question for you: `ExtrinsicContextBuilder` doesn't seem to support setting an era block hash (unless it's a genesis block) - is this intentional?

splix:
Oh, _Era_ is a bit tough part. Can't recall exact reasons for that part of code, but in short I wasn't able to find any good spec for the Eras. I spent some time trying to reverse engineer it, but it wasn't clear. 

So, since it isn't a required part, I left eras with a partial implementation and using Immortal eras for now. That's why you  don't see much support for eras, besides base implementation. Hope to finish it in later releases, when it will become more clear how exactly it supposed to work. Maybe it works already, but it's not guarantied, so I don't mention it in docs, wrappers, builders, etc.

For the implementation I also took a look at the Substrate code: https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/generic/era.rs#L114 Maybe I'm missing something, but it's encoded in the same order as in my code, no?

mmagician:
@splix I agree, there aren't good specs yet regarding Era, I'll see if I can follow up on this with an appropriate team and hopefully provide a good reference for the future.

I checked your code against the Rust implementation and as you say, they seem to be in sync - my bad! Nice job reverse engineering there 👍 

Also, thanks for updating the examples, they run fine now. I should be able to complete the assessment of this milestone soon.

mmagician:
@splix One more thing that occurred to me while looking at some of the older parts of the code: you're probably aware that [redomination](https://wiki.polkadot.network/docs/en/redenomination) took place on August 21st, changing the value of 1 DOT from 1e12 Plancks to 1e10 Plancks. 

I believe this will require some updates to the `polkaj-common-types` module. While this won't stop me from accepting the 3rd milestone (since technically we've already approved the part covering common-types), it's something to keep in mind. Perhaps the next time you're having a look at the code you could tackle this -> for now I'll open an issue in your repo so that any potential users are aware.

mmagician:
@splix We actually have the specs for Era: check out Definition 6, as well as Algorithms 1 & 2 of section 1.3 from [this](https://github.com/w3f/polkadot-spec/blob/master/runtime-spec/polkadot-runtime-spec.pdf) document.

Noc2:
Hi @splix I just merged @mmagician [notes](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/java-client_3_mmagician.md). You milestone is now accepted. I will forward your invoice internally.

splix:
@mmagician thank you for for the link to the document. I thing I did check it out before, but somehow missed this part. Now it's very useful, and I can verify the implementation. Also thanks for reminding about DOT denomination, I saw that, but didn't realize I may have wrong formatting now. It's good idea to update it.

@Noc2 thank you.

RouvenP:
hi @splix we just sent a test transaction. Could you please confirm if received?

splix:
@RouvenP hi. Yes, I confirm that

RouvenP:
@splix thanks for confirming. Payment is sent.

splix:
@RouvenP thank you, I confirm receiving

PR #36, opened 2020-09-16
reuse the guidelines from General Grants Program as the template

mmagician:


Noc2:
I would change it a little bit, since for the delivery repo, it's a template for the submission and no guidelines. Maybe it’s easier to change guidelines of the general grants program also to simply suggest using the template. I have the feeling a template might lead to better submission compared to guidelines.   

We probably should also integrate the invoice form into the general grants program in parallel. Maybe we can work first on two branches (general + open grants repos) in parallel.

Noc2:
@mmagician I integrated your changes here: https://github.com/w3f/General-Grants-Program/pull/373 and instead of integrating it into the template. I would simply link in the template to the milestone delivery guidelines.

Noc2:
And PR  #66 is to update the Grant-Milestone-Delivery and include the guidelines. Let me know if you think this looks good. In this case I would close this PR and merge the others. But feel free to also change it.

PR #41, opened 2020-10-03
external evaluation for lunie-governance

blrhc:


mmagician:
@blrhc Have you actually voted on any proposals?

PR #84, opened 2021-01-18
submit KSM-embeddable-tip-or-donate-button evaluation

aleixosk:
Submit evaluation of KSM tip button milestones 1 & 2.

### DIFF
--- a/evaluations/KSM-embeddable-tip-or-donate-button_1_aleixosk.md
+++ b/evaluations/KSM-embeddable-tip-or-donate-button_1_aleixosk.md
@@ -1,0 +1,29 @@
+# Evaluation
+
+
+* **Status:** In Progress
+* **PR Link:** [https://github.com/w3f/Open-Grants-Program/pull/64](https://github.com/w3f/Open-Grants-Program/pull/64)
+* **Milestone:** 1
+* **Kusama Identity:** [EdAnXzJGJcfNATWahgMgewkSA84nGHZwANpVzGVXtDYetsw](https://polkascan.io/kusama/account/EdAnXzJGJcfNATWahgMgewkSA84nGHZwANpVzGVXtDYetsw) (no identity yet)
+* **Previously successfully merged evaluation:** none
+
+| Number | Deliverable | Accepted | Link | Evaluation Notes |
+| ------------- | ------------- | ------------- | ------------- |------------- |
+| 1. | Embedded code | <ul><li>[ ] </li></ul> |?| No URL parsing as proposed is done as far as I can tell |
+| 2. | Connecting to PolkadotJS extension | <ul><li>[ ] </li></ul> |[Connect.svelte](https://github.com/Shard-Labs/kusama-tips-widget/blob/e2e1629ba77449c4141d5a7c1374cbd3874ba623/src/components/modals/Connect.svelte)| Works like a charm, see UX suggestion below |
+
+## General Notes
+
+Apache 2.0 LICENSE file in the repo.
+
+### URL Parsing
+
+Perhaps I'm missing something, but I can't see this being done anywhere in the code.
+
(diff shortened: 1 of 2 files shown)


Noc2:
Thanks for the evaluation. It looks good and I will merge it. It would be great if you could update it again, once the team integrated some of your proposed changes.

dark64:
Hello! @aleixosk 

Thanks for the great review and suggestions! We made the following improvements:

Milestone 1:

- Changed base URL to sanitized full URL (as described in the proposal). After some examination of the different possibilities and effort required to restrict proposals to only one tip per specific URL, we decided this should not be strictly required. If you think this is should be required, we could discuss this further.
- If polkadot{.js} extension is installed, the landing page with installation instruction is no longer shown.

Milestone 2:

- We made some changes to the description displayed in the Propose Tip tab regarding the bonds. It is now more clear that the bond is required only for the public tips (users who are not part of the Council). We decided to not calculate the bond with the fixed values provided in the evaluation as we do not know if this is subject to change, which may cause wrong data shown to the user in the future if the values do change. We tried to use [https://polkadot.js.org/apps/#/treasury/tips](https://polkadot.js.org/apps/#/treasury/tips) as a reference but could find any info about this.
- We made sure the button is disabled if the entered amount surpassed the available balance. Also, we fixed some validation bugs that would still render a fee when previously entered input is erased.
- We added comments to the code where necessary and updated the documentation with more information about available properties that can be passed to the widget explaining each of them with some code examples.
- As we used Svelte to build this plugin with their `getContext` API to share data between the components, we realized this isn't easily testable with unit tests as `getContext` and `setContext `API's are expected to be called as part of a render cycle of a component tree, which you can’t easily fake. Therefore, we made sure to test the widget thoroughly, though unit tests were not mentioned in the original proposal as part of the deliverables.

Let us know if you have any questions.
Thanks!

aleixosk:
Hi @dark64. Many thanks for the swift response and for the improvements. I've submitted the finished evaluation [here](https://github.com/w3f/Grant-Milestone-Delivery/pull/88). 

### URL sanitisation

I have checked with @Noc2 and we are fine without the restriction to one tip per URL - I agree that the effort required to implement it is probably disproportional to the added benefit. However, the current implementation of `sanitizeUrl` fails to recognise parameter values containing caps, hyphens and other characters. I went ahead and created a [pull request](https://github.com/Shard-Labs/kusama-tips-widget/pull/2) that fixes this, which I see you've already merged - thanks!

### Propose Tip description

Looks good to me. My bad - I should have specified where the constants for the bond calculation were, and you're right that they may change in the future. They are accessible from the API and I have created another [pull request](https://github.com/Shard-Labs/kusama-tips-widget/pull/3) implementing this feature. It's a pity that there is no API method that does this atm, but I think it's solid enough like this. Only if the approach used to calculate the bond is modified will it ever return erroneous values, but not if the constants are. This is not part of the evaluation, just a personal contribution. Also here, thanks for the quick merge.

Something @Swader mentioned in the [application PR](https://github.com/w3f/Open-Grants-Program/pull/64) was highlighting the possibility that the bond may be slashed. Personally, I think it's fine without the warning, maybe let's give him a couple of days to see if he has anything to add?

From my side, the delivery is accepted.

Swader:
I feel good about this as is 👌 it is now up to the community to test and show whether or not the concept of tips is clear enough (from experience, most expect to be tipping from their own account).

anamst:
Hi @alxs we are checking to see if there is anything pending from our side that needs to be changed.  We are doing some housecleaning and would like to finalize this project and start focusing on other stuff. Could you advise on how to proceed further, do we need to clear some paperwork or do something else to finalize the grant? Thanks

Noc2:
Hi @anamst the grant is accepted and the invoice is forwarded internally, see https://github.com/w3f/Grant-Milestone-Delivery/pull/74#issuecomment-766347988.

anamst:
Thanks for the quick response!

PR #88, opened 2021-01-22
finish KSM tip button evaluation

aleixosk:


Noc2:
Thanks again for the evaluation and updating it. I will check internally how much KSM we will pay you for it, since the price went up quite a bit. But we will definitely pay you for the evaluation. 

Let me know in case there is an issue with the KSM address, which you provided.

PR #99, opened 2021-02-11
 SubsCrypt milestone 1

SaberDoTcodeR:

### DIFF
--- a/deliveries/SubsCrypt-1.md
+++ b/deliveries/SubsCrypt-1.md
@@ -13,5 +13,5 @@ Please provide a list of all deliverables of the milestone extracted from the in
 | ------------- | ------------- | ------------- |------------- |
 | 1. | Contract Implementation |[Solidty](https://github.com/oxydev/SubsCrypt-solidty) - [Ink!](https://github.com/oxydev/SubsCrypt-ink)| Contract is implemented in two language which can be used both in EVM pallete and Ink! platform.| 
 | 2. 3.  | Article & Design Front-End component | [Git Version](https://github.com/oxydev/SubsCrypt-docs) - [Web Version](https://oxydev.github.io/SubsCrypt-docs/)| SubsCrypt Documntaion platform that more details will be written in next phases.| 
-| 4.  | White-Paper | [PDF]()| SubsCrypt platform and future plans are explained in an abstract level .| 
-| 5.  | Polkadot Standards Proposals (PSPs) | [Proposal](#)| Our proposal for implementing the Subscription mechanism with refund method.| 
+| 4.  | White-Paper | [PDF](https://github.com/oxydev/SubsCrypt-docs/blob/master/SubsCrypt-White-Paper.pdf)| SubsCrypt platform and future plans are explained in an abstract level.| 
+| 5.  | Polkadot Standards Proposals (PSPs) | [Proposal](https://github.com/w3f/PSPs/pull/12)| Our proposal for implementing the Subscription mechanism with refund method.| 
--- a/deliveries/SubsCrypt-1.md
+++ b/deliveries/SubsCrypt-1.md
@@ -1,0 +1,17 @@
+# Milestone Delivery :mailbox:
+
+> Only the GitHub account, which is responsible for the pull request of the accepted application is allowed to submit milestones. Don't remove any of the mandatory parts presented in bold letters or as headlines!
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).**  
+
+* **PR Link:** [171](https://github.com/w3f/Open-Grants-Program/pull/171). 
+* **Milestone Number:** [1](https://github.com/w3f/Open-Grants-Program/blob/master/applications/SubsCrypt.md#milestone-1--smart-contract)
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | Contract Implementation |[Solidty](https://github.com/oxydev/SubsCrypt-solidty) - [Ink!](https://github.com/oxydev/SubsCrypt-ink)| Contract is implemented in two language which can be used both in EVM pallete and Ink! platform.| 
+| 2. 3.  | Article & Design Front-End component | [Git Version](https://github.com/oxydev/SubsCrypt-docs) - [Web Version](https://oxydev.github.io/SubsCrypt-docs/)| SubsCrypt Documntaion platform that more details will be written in next phases.| 
+| 4.  | White-Paper | [PDF]()| SubsCrypt platform and future plans are explained in an abstract level .| 
+| 5.  | Polkadot Standards Proposals (PSPs) | [Proposal](#)| Our proposal for implementing the Subscription mechanism with refund method.| 


mmagician:
Thanks, we'll look into it as soon as possible, though we ask for patience :)

mmagician:
@SaberDoTcodeR Thanks for your delivery. I would like to ask you to revise the submission document to reflect all individual points & links from the original application since some are missing at the moment. In particular:
1. > we will implement the contract and deploy it to a test net.

Info missing

2. > full documentation of API that explains how to connect to and use it.

Missing

3. > Testing Guide

I see you have tests in your ink! contracts but without any instructions on how to use it

4. > technical specification of this project
 
Your whitepaper points to the [online docs](https://oxydev.github.io/SubsCrypt-docs/#/), but there I can only find the same information as in the white paper, there are no technical details.

5. > design UI and UX of front-end component with adobe xd.

In the online docs I could only find 2 screenshots of the dashboard and a popup. Could you provide us links/access to the full mock ups of your front end please?

SaberDoTcodeR:
Hi @mmagician , Thanks for your consideration. 
1. > we will implement the contract and deploy it to a test net.
we have built and generated the wasm file of the contract and we have tried to deploy it on the test net of edgeware(beresheet) and there were some errors that we are working on to fix with their help. So I think it's better that we continue to the next milestones and we will work on fixing this deployment problem ASAP.

2. > full documentation of API that explains how to connect to and use it.
3. > Testing Guide
4. > technical specification of this project
5. > design UI and UX of front-end component with adobe xd
Well, we have updated our documentation repo and added more explanations on the technical facet.

mmagician:
Thanks for adding more info.

After I tried following your testing instructions, I ran into the following error when running `cargo +nightly test` (with newest nightly):
```
error[E0034]: multiple applicable items in scope
   --> /Users/marcin/.cargo/registry/src/github.com-1ecc6299db9ec823/bitvec-0.20.1/src/order.rs:291:15
    |
291 |         if ct == R::BITS {
    |                     ^^^^ multiple `BITS` found
```

---

> So I think it's better that we continue to the next milestones and we will work on fixing this deployment problem ASAP.

So do you prefer to wait with the further evaluation of this milestone until you have the next one ready? Please clarify as I didn't fully get your point.

SaberDoTcodeR:
> After I tried following your testing instructions, I ran into the following error when running cargo +nightly test (with newest nightly):

Well, I checked the tests right now, and I had no problem with these configurations:
IMAGE 2021-02-19 16:16:31

> So do you prefer to wait with the further evaluation of this milestone until you have the next one ready? Please clarify as I didn't fully get your point.

We'd rather receive the grant of this milestone while we are doing the next two milestones simultaneously because those two milestones require much more resources for development. So if it's feasible, we'd be glad that you finish the evaluation of this milestone, therefore, the development of the next two phases won't get blocked due to a lack of financial recourse.

> Lastly, please add some READMEs to your repos (it's fine if they link to your online docs, just as long as it's not completely empty). Thanks!

Done.

mmagician:
@SaberDoTcodeR Can you check which rust version you're using? As I mentioned, I tried with latest nightly`rustc 1.52.0-nightly (9b471a3f5 2021-02-19)` and it still fails on `cargo +nightly test`.

SaberDoTcodeR:
I updated my nightly version to yours:
`  nightly-x86_64-unknown-linux-gnu updated - rustc 1.52.0-nightly (9b471a3f5 2021-02-19) (from rustc 1.51.0-nightly (04caa632d 2021-01-30))`
 but I didn't face any error, maybe if ,as it's mentioned in the issue of bitvec, you add `funty = "=1.1.0"` to your `Cargo.toml` file, the problem would be solved.

SaberDoTcodeR:
> @SaberDoTcodeR Can you check which rust version you're using? As I mentioned, I tried with latest nightly`rustc 1.52.0-nightly (9b471a3f5 2021-02-19)` and it still fails on `cargo +nightly test`.

I added a temporary fix to the `Cargo.toml` that will solve your problem, please pull the master of [SubsCrypt-ink repo](https://github.com/oxydev/SubsCrypt-ink) for changes.

mmagician:
@SaberDoTcodeR That worked. Not sure why you call it a temporary fix?

Also, there are a multitude of warnings generated when building & testing your code, as well as a result of running `cargo clippy`. I would like to ask you to address these please before proceeding.

SaberDoTcodeR:
> @SaberDoTcodeR That worked. Not sure why you call it a temporary fix?

Because this is a conflict between two dep libraries (funty and bitvec) , so I forced the Cargo to use an older funty version. When they fix it, I will remove that line so it will use the latest version of funty too.

> Also, there are a multitude of warnings generated when building & testing your code, as well as a result of running cargo clippy. I would like to ask you to address these please before proceeding.

I fixed those warnings, thanks for mentioning that.

mmagician:
That might be, but the tests are now failing.

SaberDoTcodeR:
Sorry for the inconvenience, it should be working now.

mmagician:
@SaberDoTcodeR Apologies for taking so long. You'll find my notes below:

1. Is it necessary to include the `crates` folder in your `ink!` repo? If you want to pin the dependencies, as far as I know you should be able to specify the github repo's concrete version/tag/branch, see for example [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)

2. Most of the method documentation is redundant, e.g.
```
        /// * duration : new duration of plan [in days? seconds?]
        /// * prices : vector that contains "price" amounts
        /// * plan_index : index of plan
```
3. Struct that represents a payment admission is named `Object` - this will be confusing.
4. Difference between `SubscriptionRecord::subscription_time` vs `SubsCrypt::start_time`?
5. `index_to_address: HashMap<u128, Account> `
doesn't make much sense IMO, why not use an array and access by index instead? Presumably the indices will be incremental
6. On line 204:
-> arrays provided could be of different sizes. There should be a check at the beginning to verify all the inputs

7. Use `add_plan` logic from within `provider_register` logic to avoid code duplication, or refactor for simplicity.
8. Violation of "check first, write last" rule: `subscribe` method will create a user even if a provider doesn't exist
9. Lines 306-314: if you reach this point, you know from the previous assert that a PlanRecord must have just been created, so rather than accessing it again on 314, refactor the code to make sure you already have the reference.
10. In addition, the "if" statement is not needed (again due to your previous assert and the check inside `check_subscription`)
11. Lines 319-325: you are re-using all the fields from `consts`, why not just clone it?
12. Can you explain the logic behind:
```
assert_eq!(self.transfer(*addr, consts.price * (1000 - consts.max_refund_percent_policy) / 1000),Ok(()))
```
for percentages, shouldn't it be 100? Either way it would be great to get more info on this, preferably both inline as well as in your official docs

13. The following lines (339-340) are redundant IMO, as these variables are only accessed once:
```
let dur = consts.duration;
let max_percent= consts.max_refund_percent_policy;
```

14. Please explain the following logic in your docs & ideally inline code comments too:
```
assert!(time_percent <= 1000);
if 1000 - time_percent > record.plan.max_refund_percent_policy {
    time_percent = record.plan.max_refund_percent_policy;
} else {
    time_percent = 1000 - time_percent;
}
```

15. `if time_percent < record.plan.max_refund_percent_policy {`
is confusing, especially since you've recently re-assigned the value of `time_percent` - could you perhaps integrate it into the previous "if" statement?

16. The current test suite is fairly extensive, but only covers the "positive", i.e. expected scenarios.
There aren't any tests for edge cases (e.g. supplying invalid address, empty vectors, wrong plan index etc.)

SaberDoTcodeR:
Hi @mmagician, Thanks for your review. 

> 1.Is it necessary to include the crates folder in your ink! repo? 

This was sth that we had difficulties handling, so I hardcoded some of our deps libraries to avoid the problem of building and testing our code.

> 2. , 3. , 6. , 7. , 8. , 11.

We review our comments and code and solve and refactor these mentioned problems.

>4. Difference between SubscriptionRecord::subscription_time vs SubsCrypt::start_time?

The latter one is the beginning of the contract initiated time, it's used in the linked list structure, and the former one is about when a person joined a plan. Therefore they are totally different.

>5. index_to_address: HashMap<u128, Account>
doesn't make much sense IMO, why not use an array and access by index instead? Presumably the indices will be incremental

We discussed this in our team and for the sake of better performance and easier implementation we choose the way.

>9. , 13.

You are right about these two, but due to the `clippy` warning about mutable and immutable borrow, we had no other choice.

>10. In addition, the "if" statement is not needed (again due to your previous assert and the check inside check_subscription)

If you check out this function `check_subscription`, it can be observed that there are several conditions that could trigger the false return, so we have to put that `if`.


>12. , 14. , 15.

We refactor this part of the code for better code readability, I added some inline comments and You can also check out this part of [our docs](https://oxydev.github.io/SubsCrypt-docs/#/introduction/trust?id=refund) for more explanation, and I also have to mention that those `1000` are for more precision. For example, 1% will be 10 on our scale.

>16. The current test suite is fairly extensive, but only covers the "positive", i.e. expected scenarios.
There aren't any tests for edge cases (e.g. supplying invalid address, empty vectors, wrong plan index etc.)

we added some more marginal tests that cause the code to panic for better code coverage. Those test have         `#[should_panic]` attribute.

mmagician:
@SaberDoTcodeR Thank you for addressing my comments, I'll review your updates in the coming days.

mmagician:
@SaberDoTcodeR here are my follow-up comments:

---
Regarding crates directory: please fix this.
I just tried using the newest ink! tag `version = "3.0.0-rc3"`, removing the crates folder and it worked just fine.
If you are not comfortable working with rust dependencies, I suggest investing more time into this. Including the libraries as part of your repo is not good coding practice and will not be accepted as part of a milestone delivery for the Web3 Foundation.

---
> We discussed this in our team and for the sake of better performance and easier implementation we choose the way.

Could you please outline your reasoning here? I struggle to see any benefits of using a HashMap when an array would suffice.

---
After looking more closely at the `subscribe` method, I discovered that your code is more inefficient than I initially thought.
First of all, on line 303 you are getting the length of `plan_record.subscription_records` which at that point is always 0, since you've just initialised it.
Second, it's the accessing of `plan_record` via `self.records` (what I mentioned earlier), which should be refactored, since you just created the variable, so you shouldn't need to get a reference by accessing `records`
Third, as far as I can see, `let record: SubscriptionRecord` doesn't depend on any previous variables. So rather than creating a record and pushing it to `plan_record`'s array, why not init `plan_record` with it already?
Lastly, perhaps most crucially: I couldn't find any place in your code where the `plan_record.subscription_records` would have more than one entry. It seems that the only place that you push elements to that array is there at line 313 - and that's always going to be one element. This seems like a design flaw, but please correct me if I'm wrong, I might have missed some reference or function. Either way, this needs a major refactor given the above points.

---
> You are right about these two, but due to the clippy warning about mutable and immutable borrow, we had no other choice.

Again, this will require some research from your side, as it it looks like you don't know why clippy gives you these warnings. As a tip from my side, let me suggest that `let consts: PlanConsts` needs not be mutable in your code.

---
> If you check out this function check_subscription, it can be observed that there are several conditions that could trigger the false return, so we have to put that if.

Let's walk through a scenario. Assume a record exists, i.e. `self.records.contains_key(&(user, provider_address))` evaluates to `true`

Then `check_subscription` wouldn't return false, because it checks for NON-existance of the record: `!self.records.contains_key`. Instead, we can assume that return of `check_subscription` is `true`

Consequently, the `assert` will trigger the error, as it's expecting a `false` return statement. Therefore, execution will stop there and not reach the `if` statement. Am I correct in my reasoning?

Second scenario is where a record doesn't exist. `check_subscription` evaluates to `false`, `assert` passes and we reach the **exact same** check in the `if` statement, which will evaluate to `true` and code inside the `if` will be executed.
Therefore, I don't see a reason for having that statement.

---
> Violation of "check first, write last" (again)

I see you've just followed the exact point that I mentioned regarding not creating a User when a Provider doesn't exist. But you're still doing many checks AFTER the user is created in the `subscribe` method (e.g. `assert_eq!(consts.price, value, "You have to pay exact plan price");`), which don't depend on the user creation. 

---
Given the high number of bad coding practices and other code parts which demonstrate that you're doing only doing what is minimally required, I would strongly recommend that you take your time to polish the codebase.

SaberDoTcodeR:
@mmagician Sorry for the delay. As you rightly mentioned that we needed to polish our codebase, we did a thorough review of the codebase and our repository structure. Please check them out, it will be much easier for you to review the code as well! Here are some points I think need explaining: 

> Let's walk through a scenario where a record exists, i.e. self.records.contains_key(&(user, provider_address)) evaluates to true

So each user can subscribe to multiple plans of a provider, for instance: we have Alice as a user and a dairy company as a provider. The dairy company has two kinds of plans:

 1. Monthly milk delivery (plan index = 0)
 2. Monthly butter delivery (plan index = 1)

Now if Alice wants, she can subscribe to both of these plans and have a valid subscription, So `!self.records.contains_key(&(caller, provider_address))` can be either false or true (depending upon her choice). 

If Alice subscribes for the first time ever to plans, assuming she selects the first plan (plan index = 0) of the dairy company, then we will enter the IF statement in ` fn subscribe` [here](https://github.com/oxydev/SubsCrypt-ink/blob/main/src/lib.rs#L420); But when Alice tries to subscribe to another plan (plan index = 1) of the dairy company then we will go in else statement [here](https://github.com/oxydev/SubsCrypt-ink/blob/main/src/lib.rs#L441). So here you can see that this IF statement is necessary for subscript.

mmagician:
@SaberDoTcodeR Thanks for the update. I'll review your changes soon!

mmagician:
@SaberDoTcodeR I've finished reviewing your updates, overall I can see that you've put more effort into this now and the quality is looking better.
I've got only a few minor requests:
1. I've fixed some typos in your docs, see [PR](https://github.com/oxydev/SubsCrypt-docs/pull/1)
2. I've proposed a cleaner way of verifying the presence of an element in your HashMap and then accessing it. I fixed it in some places, and I left TODOs in others for you to do the same there. While the code you have currently is not wrong and it works, I believe the new proposed way is cleaner, more concise and more "rusty". I actually got a second opinion on this to make sure I'm not biased & limited by my own knowledge, and my colleague agreed with my line of thinking on this one (still could be subjective, but now we have a sample size of 2!)
3. It would make reading your code cleaner had you used [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html) in a couple of places, or bundle the functions' return data into struct definitions. You'll find TODOs for these as well.
4. Minor refactoring.
3. Some unit tests should probably be named more intuitively, e.g. `refund_works2` should be `refund_doesnt_work`, as that's what it's actually checking. This is just cosmetic. 
4. I'd suggest to remove the `deploy` folder. If anything, you can place these build files as part of your release, but I think it's enough to just have instructions on how to build them (as you already do)

For points 2-4, you'll find my suggestions along with TODOs in [this PR](https://github.com/oxydev/SubsCrypt-ink/pull/13). Let me know what you think. If that makes sense to you, these should be fairly quick fixes on your side and then we can move forward with the future milestones. Congrats once again on the progress!

SaberDoTcodeR:
Hey @mmagician, thanks for your complaimants. We fully resolved the points you mentioned in [this PR](https://github.com/oxydev/SubsCrypt-ink/pull/13). we also release our tag v0.1 which contains the binaries and deleted the `deploy` folder.
We are appreciating your comments and codebase refactors.

mmagician:
@SaberDoTcodeR Despite the long process and multiple issues, I'm happy to tell you that your milestone has been accepted!
You can find the full evaluation [here](https://github.com/w3f/Grant-Milestone-Delivery/pull/138). Congratulations and good luck with the front & back end parts!

PR #105, opened 2021-03-05
Commit php-scale-lib milestone 1

gmajor-encrypt:

### DIFF
--- a/deliveries/php-scale-lib_milestone_1.md
+++ b/deliveries/php-scale-lib_milestone_1.md
@@ -1,0 +1,17 @@
+# Milestone Delivery :mailbox:
+
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone.**  
+
+* **PR Link:** [php-scale-lib](https://github.com/w3f/Open-Grants-Program/pull/235)
+* **Milestone Number:** 1
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | Source code   | https://github.com/gmajor-encrypt/php-scale-codec |
+| 2. | License       | MIT license |
+| 3. | Unit Tests    | https://github.com/gmajor-encrypt/php-scale-codec/tree/master/test/Codec | make test |
+| 4. | Documentation | https://github.com/gmajor-encrypt/php-scale-codec#basic-usage |
+| 5. | Packagist     | https://packagist.org/packages/gmajor/substrate-codec-php |


gmajor-encrypt:
Why ci fail?

Noc2:
Don’t worry about it and thanks for the delivery. We will look into it as soon as possible.

gmajor-encrypt:
@alxs hello, any update?

alxs:
Hi @gmajor-encrypt, sorry for the long silence. We're currently pretty busy but I'll provide some feedback on the evaluation by the end of this week.

alxs:
@gmajor-encrypt allow for a couple more days, so far I have only positive feedback.

gmajor-encrypt:
@alxs thank you reply. I will start milestone 2 work ahead of time

alxs:
Hi @gmajor-encrypt. I'm back with some useful feedback.

Thanks a lot for all the effort you've put into  this, the delivery is great. I have a few requests in no particular order:
- May I ask you to upgrade to PHP 8? I only had to fix a few things in order to make it work so this shouldn't be much trouble.
- What's the purpose of the `CompactU32` type? As I see it `Compact` would work for all cases but maybe I'm missing something.
- In general, the library needs more tests. Specifically, a) testing for exceptions currently isn't done at all and b) there should be at least a separate test for encoding and decoding for each type, each with a set of legal and illegal values. I encourage you to also leverage [coverage annotations](https://edorian.github.io/2011-11-04-An-introduction-to-phpunits-covers-annotation/) in your code and to create generic tests that can be reused for all types of each basket wherever possible.
- Ideally, the interface exposed to the user is identical for all types, i.e. the `encode` and `decode` pattern should work everywhere instead of the current `process` outliers for `Vecs` and others. If this is not possible, it should at least be documented.
- Documentation needs to be significantly expanded in general. Almost all in-code documentation is currently auto-generated and hence redundant and doesn't help in learning what a function does. More inline comments would also be welcome. Also see the [AssemblyScript](https://github.com/LimeChain/as-scale-codec) or [Swift](https://github.com/tesseract-one/swift-scale-codec) SCALE implementations for very good examples of informative READMEs.
- One of `Bytes` and `HexBytes` is redundant.
- Optional booleans are an [exception](https://substrate.dev/docs/en/knowledgebase/advanced/codec#options) and are currently handled incorrectly.
- `ScaleDecoder` seems to also encode? Perhaps `ScaleInstance` or the like would be a more appropriate name.
- For vectors, please include some test cases using arrays of different types, as well as for the other complex data types.
- I haven't looked that closely at the `Enums` implementation but the test therefor seems to test for something else than what I was expecting. What is this based on? I'm looking at the Substrate [docs](https://substrate.dev/docs/en/knowledgebase/advanced/codec#enumerations-tagged-unions) for reference.

gmajor-encrypt:
> Hi @gmajor-encrypt. I'm back with some useful feedback.
> 
> Thanks a lot for all the effort you've put into this, the delivery is great. I have a few requests in no particular order:
> 
> * May I ask you to upgrade to PHP 8? I only had to fix a few things in order to make it work so this shouldn't be much trouble.
> * What's the purpose of the `CompactU32` type? As I see it `Compact` would work for all cases but maybe I'm missing something.
> * In general, the library needs more tests. Specifically, a) testing for exceptions currently isn't done at all and b) there should be at least a separate test for encoding and decoding for each type, each with a set of legal and illegal values. I encourage you to also leverage [coverage annotations](https://edorian.github.io/2011-11-04-An-introduction-to-phpunits-covers-annotation/) in your code and to create generic tests that can be reused for all types of each basket wherever possible.
> * Ideally, the interface exposed to the user is identical for all types, i.e. the `encode` and `decode` pattern should work everywhere instead of the current `process` outliers for `Vecs` and others. If this is not possible, it should at least be documented.
> * Documentation needs to be significantly expanded in general. Almost all in-code documentation is currently auto-generated and hence redundant and doesn't help in learning what a function does. More inline comments would also be welcome. Also see the [AssemblyScript](https://github.com/LimeChain/as-scale-codec) or [Swift](https://github.com/tesseract-one/swift-scale-codec) SCALE implementations for very good examples of informative READMEs.
> * One of `Bytes` and `HexBytes` is redundant.
> * Optional booleans are an [exception](https://substrate.dev/docs/en/knowledgebase/advanced/codec#options) and are currently handled incorrectly.
> * `ScaleDecoder` seems to also encode? Perhaps `ScaleInstance` or the like would be a more appropriate name.
> * For vectors, please include some test cases using arrays of different types, as well as for the other complex data types.
> * I haven't looked that closely at the `Enums` implementation but the test therefor seems to test for something else than what I was expecting. What is this based on? I'm looking at the Substrate [docs](https://substrate.dev/docs/en/knowledgebase/advanced/codec#enumerations-tagged-unions) for reference.

 @alxs Thanks you reply, I will make some changes as soon as possible

alxs:
Hey @gmajor-encrypt, any updates?

gmajor-encrypt:
> Hey @gmajor-encrypt, any updates?

Some things have been delayed recently, and the update will begin today. I have a question
```Ideally, the interface exposed to the user is identical for all types, ie the encode and decode pattern should work everywhere instead of the current process outliers for Vecs and others. If this is not possible, it should at least be documented.```
I do not understand the meaning of this

**Comment by alxs** [comment]
Glad to hear that. Sure, encoding is done at least in the tests as follows for most types:
``` php
$scaleBytes = new ScaleBytes(...);
$codec = $generator->Type($scaleBytes);
$value = $codec->decode();
```

But then for `Vecs` and others it becomes:
``` php
$codec = new ScaleDecoder($generator);
$value = $codec->process("Type", new ScaleBytes(...));
```

This is confusing and ideally one would be able to use the same pattern everywhere.

gmajor-encrypt:
> Glad to hear that. Sure, encoding is done at least in the tests as follows for most types:
> 
> ```
> $scaleBytes = new ScaleBytes(...);
> $codec = $generator->Type($scaleBytes);
> $value = $codec->decode();
> ```
> 
> But then for `Vecs` and others it becomes:
> 
> ```
> $codec = new ScaleDecoder($generator);
> $value = $codec->process("Type", new ScaleBytes(...));
> ```
> 
> This is confusing and ideally one would be able to use the same pattern everywhere.

The first method is to have a **pre-defined type**, and the second method is to correspond to a **custom (composite) type**.

You are right, this will cause some confusion. I will use the second method completely, which can be applied to most scenarios

gmajor-encrypt:
@alxs Hello, I have recently made some changes in accordance with the revised comments, please check again

alxs:
Thank you, I'll look into it this week.

alxs:
Hey again, many thanks for the changes. I gave it a deeper look this time and came up with the following:

- I believe there are rounding errors happening with large integers. You can check that

  ```php
  $codec->createTypeByTypeString("Compact")->encode(2 ** 536 - 1)
  ```
  also throws an `OutOfRangeException`, although this should be a legal value. This is probably due to the [limitations of `double` precision](https://stackoverflow.com/a/1848762). This is an issue especially since it means that values are likely not encoded correctly either. Perhaps you should try using [BigIntegers](https://pear.php.net/package/Math_BigInteger).
- In general I'd recommend adding tests for all edge cases (i.e. min and max values) in order to avoid precisely this sort of thing.
- Shouldn't it be possible to decode compact integers without specifying the type? Otherwise one would need to manually check their length before doing so, which is extra overhead that this type is supposed to eliminate. Besides, this should be easy to check based on the number of bytes or on the two least significant bits in the first one. If possible I'd suggest removing the `Compact<type>` pattern altogether.
- Same goes for Vectors. Two things here:
  1. The length should be decoded as a generic compact integer and not in fixed four-byte mode.
  2. The type should be inferred from the number of bytes divided by the length instead of being manually provided.
- The first item above also applies to the length of strings. Also these are currently encoded in ASCII instead of UTF-8, I think what you're looking for is [`utf8_encode`](https://www.php.net/manual/en/function.utf8-encode.php) (I'm not entirely sure about this, have a look at the comments under the link and in doubt add some [tests with special characters](https://github.com/paritytech/parity-scale-codec/blob/master/src/codec.rs#L1331) to make sure).
- Tuples are missing. I'm not sure if that's what [this encoding](https://github.com/gmajor-encrypt/php-scale-codec/blob/master/test/Codec/TypeTest.php#L69) is intended to represent, but Vectors containing values of different types seem to be [unsupported](https://substrate.dev/docs/en/knowledgebase/advanced/codec#vectors-lists-series-sets) and the encoding for tuples works differently. I'm inclined to think this functionality should be removed from Vectors altogether since other implementations won't be able to interpret it, but maybe I'm missing something.
- Structs are nicely implemented. This is probably related to the previously mentioned issue with these, but how can I create a struct consisting of generic compact integers?
- Enums aren't implemented [according to specification](https://substrate.dev/docs/en/knowledgebase/advanced/codec#enumerations-tagged-unions).
- `I128` is missing. Are there any hidden complexities in implementing this?
- I see that [Results](https://substrate.dev/docs/en/knowledgebase/advanced/codec#results) aren't implemented, nor are they mentioned in the application. If you want to implement them as well, feel free to  submit an amendment for M2 and increase the price a little.

I've pinged @Swader about the delivery and he'll provide some feedback from his side soon. In the meantime, feel free to start looking into the above. Pending his feedback and if everything looks good after this, I'll be happy to accept the milestone.

gmajor-encrypt:
@alxs Thank you reply. I will make some changes as soon as possible to make it more perfect

gmajor-encrypt:
@alxs hey, I have recently made some changes in accordance with the revised comments, please check again.
In addition, I am going to implement the [Results](https://substrate.dev/docs/en/knowledgebase/advanced/codec#results) type in m2

freehere107:
Regarding Tuples, he implemented it here https://github.com/gmajor-encrypt/php-scale-codec/blob/37e1b9a01fac0e859bdaf63fcf94b665ca918cd8/src/Codec/Types/ScaleInstance.php#L95
and 
https://github.com/gmajor-encrypt/php-scale-codec/blob/37e1b9a01fac0e859bdaf63fcf94b665ca918cd8/test/Codec/TypeTest.php#L177

gmajor-encrypt:
Regarding Tuples, he implemented it here https://github.com/gmajor-encrypt/php-scale-codec/blob/37e1b9a01fac0e859bdaf63fcf94b665ca918cd8/src/Codec/Types/ScaleInstance.php#L95
AND
https://github.com/gmajor-encrypt/php-scale-codec/blob/37e1b9a01fac0e859bdaf63fcf94b665ca918cd8/test/Codec/TypeTest.php#L177

alxs:
Thanks for the changes @gmajor-encrypt. The updated Compact type and the Tuples implementation look neat, though there are still a few issues with regards to my comments above:

- Big integers fail above `2**535-1`, instead of `2**536-1`.
- Decoding doesn't seem to work for big integers with a number of bytes that is not defined as an unsigned type. The following throws an error:

  ```php
  $codec->process("Compact", new ScaleBytes("0300000040"));
  ```
- Decoding large Compact integers should return GMP integers for large values (potentially for all of them in order to avoid type errors, but I leave this up to you) or otherwise the same imprecisions will arise when decoding. Encoding and decoding tests for all edge cases would really be immensely helpful here.
- Vector types should be inferred instead of being provided manually and vectors containing values of different types shouldn't be supported. But let me know if these are indeed supported in the Substrate implementation, we're aware that the documentation may be a bit outdated.
- The README (which, by the way, looks great) needs to be updated to reflect the new Compact, Enum, Tuple and Vec implementation.

gmajor-encrypt:
@alxs , thank you for your reply

For vector, I searched https://substrate.dev/docs/en/knowledgebase/advanced/codec#vectors-lists-series-sets,

https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/Vec.spec.ts

https://github.com/itering/scale.go/blob/2de1df39be063a013bfd3b9d8641fde5a7609735/types/types.go#L200

The sub type needs to be provided manually, which is the key parameter of decode/encode.
Maybe I understand something wrong, please point it out, thank you

alxs:
I see, thanks for the input! Parity's should be the reference implementation here. Agree regarding providing the subtype manually, but the values should be same-typed as you can see [here](https://github.com/paritytech/parity-scale-codec/blob/master/src/generic_array.rs#L19). This is also specified in the documentation you linked to. It seems those other implementations don't fully conform to the specification.

gmajor-encrypt:
@alxs hey, I have made some changes, and some issues have been mentioned above.
Corresponding to ```Vector```, I can't find anything that needs to be modified, and further communication may be needed here

alxs:
Thanks @gmajor-encrypt, I'll look into it in the coming days. @Swader wanna take a look at this now?

Regarding vectors, could you send me a message to `@aleixo:web3.foundation` on Element?

gmajor-encrypt:
@Swader @alxs  hey, can help me take a look at this

alxs:
Hey @gmajor-encrypt. Sorry for the delay, I was under the impression that you would get back to me on Element. @Swader probably won't have time to review your delivery but I have some last bits of feedback I already mentioned:
- Compact should fail above `2**536-1`, instead of `2**535-1`.
- Could you please specify in the documentation when Compact returns which type.
- The README needs to be updated to reflect the new (and correct) Enum implementation. Besides, it's likely not entirely correct since in the following test
  
  ```php
  $codec = $codec->createTypeByTypeString("Enum");
  $codec->typeStruct = ["int" => "u8", "bool" => "bool"];
  $codec->init(new ScaleBytes("0x002a"));
  $this->assertEquals(42, $codec->decode());
  $this->assertEquals("0101", $codec->encode(["bool" => true]));
  ```
  I believe `$codec->decode()` should return `["int" => 42]`.

And feedback provided via Element:
- `Vec<T>`: length is `Compact<u32>`, **not** `Compact` (I believe you had this right initially, sorry!)
- `Tuple`: max number of values is 18

alxs:
Thanks for addressing the last few points. I'm happy to tell you that the milestone is a pass: you can find my evaluation notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/php-scale-lib_1_alxs.md). I'll forward your invoice internally.

Please keep us posted in case you run into any issues with M2, and good luck!

fededubbi:
Good morning, 

Could you please specify your complete address on the invoice?
It is not complete.

Many thanks,
Federica

gmajor-encrypt:
@alxs Thank you for your patience evaluation

gmajor-encrypt:
> Good morning,
> 
> Could you please specify your complete address on the invoice?
> It is not complete.
> 
> Many thanks,
> Federica

@fededubbi The previous form can no longer be submitted, how do I need to submit another one?

Noc2:
@gmajor-encrypt You can also send the invoice directly to invoices@web3.foundation

gmajor-encrypt:
> @gmajor-encrypt You can also send the invoice directly to [invoices@web3.foundation](mailto:invoices@web3.foundation)

多谢

gmajor-encrypt:
@fededubbi hey, I have sent invoices via invoices@web3.foundation, please check

fededubbi:
All good, thanks.

Federica

RouvenP:
hi @gmajor-encrypt we transferred the payment.

gmajor-encrypt:
@RouvenP Thank you ~

PR #108, opened 2021-03-05
Commit curve_amm-milestone_1

mikolajsobolewski:

### DIFF
--- a/deliveries/curve_amm-milestone_1.md
+++ b/deliveries/curve_amm-milestone_1.md
@@ -4,7 +4,7 @@
 
 **The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).**  
 
-* **PR Link:** Please, provide a link to the initial accepted pull request of your application to the [Web3 Foundation Open Grants Program repository](https://github.com/w3f/Open-Grants-Program/pull/225). 
+* **PR Link:** https://github.com/equilibrium-eosdt/Grant-Milestone-Delivery/blob/master/deliveries/curve_amm-milestone_1.md. 
 * **Milestone Number:** 1
 
 Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
--- a/deliveries/curve_amm-milestone_1.md
+++ b/deliveries/curve_amm-milestone_1.md
@@ -1,0 +1,20 @@
+# Milestone Delivery :mailbox:
+
+> Only the GitHub account, which is responsible for the pull request of the accepted application is allowed to submit milestones. Don't remove any of the mandatory parts presented in bold letters or as headlines!
+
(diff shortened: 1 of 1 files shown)


semuelle:

alxs:
Hi @mikolajsobolewski. Sincere apologies for the delay, we've been pretty busy for the past few weeks. Let's get this going, I've had a good look today and tbh I don't think any changes will be necessary. This is very good work.

The only thing really missing is a basic tutorial explaining how to integrate your pallet as per deliverable 0b of your application. However, if you prefer to wait until M2 to submit this, that is fine by me since the deliverable is part of both milestones anyway. It also probably makes more sense.

I'd also like to just quickly ping @iamdefinitelyahuman to confirm that he or someone else from the team has also looked into this.

michwill:
Hey guys.

By "supervise" we more meant support when you ask question about math, for example, or understanding our Vyper or Python code. We don't have experience with Rust to check the code, so us looking at the _code_ would be not helpful at all.

What you can do though, you can look at our test suite, deploy functionally identical code, do 1:1 comparison by doing same calls to both (substrate and ethereum) pools and comparing that results are exactly the same. That'd be a good approach.

Another idea is that you could match the code against market data using that as a library written in Rust (is that possible?), to make it working like a really fast simulator.

And also, of course, how exactly it will be launched (is it under Curve brand or different) is to be discussed (again, given that it's a bit out of scope of our experience).

alxs:
Hey @michwill, thank you for your comments.

Re supervision, I just assumed you'd be working together but that makes sense.

As for more comprehensive testing, we would be happy to support any of the two options. This implementation is pretty much functionally identical as far as I could tell so the first approach should be doable. The current delivery is perfectly fine in the context of this milestone, but you can always submit an amendment that encompasses one of the two approaches @mikolajsobolewski.

What would be even more interesting would be the development of a tool that allows for Brownie tests or EVM unit tests in general to be run on pallet implementation equivalents of Ethereum smart contracts, as this is very likely to prove useful for other projects. That would of course constitute a larger project, but I think you'd be a good candidate to build it given your track record so let me know if you're interested. Otherwise we'll publish an RFP.

alxs:
Regarding the evaluation @mikolajsobolewski, could you add a commit hash or tag to the links in the delivery? They don't point to the correct lines any longer.

Besides, is there any way to compile the docs so that the embedded LaTeX blocks are properly rendered? Otherwise I would appreciate it if you could adapt it to a more human-readable format, or even remove them since you can always link to Curve's docs for the maths. I appreciate the effort that went into also synthesising the theory though.

mikolajsobolewski:
I'm very sorry, this file was deleted from our repository by error.  I have restored it. Link with commit hash is https://github.com/equilibrium-eosdt/Grant-Milestone-Delivery/blob/ce2d858bb344eb4356fb9e6f1f183967e39c30a5/deliveries/curve_amm-milestone_1.md 

About your second question:

1. We've ported tests from here: https://github.com/curvefi/curve-contract/tree/master/tests/pools/common/unitary and compared results when running them, results turned out to be identical. We didn't port ETH-specific tests, only those which make sense in Polkadot. 

2. We will remove the LaTeX from the code and provide a separate link to rendered version of it.

alxs:
It hasn't been deleted from the PR, so no worries. What I meant was that the links in the deliverables don't point to the correct lines in `pallets/equilibrium-curve-amm/src/lib.rs` since the file has been updated. Could you please add the commit hash after you've removed the LaTeX docs to those, and update the line numbers?

Regarding testing, that's great, I leave any further tests at your discretion.

If you can't update the file here, just close the PR and create a new one referencing it. Also note that the PR for milestone 2 now contains both files. You'll need to create a separate branch for each delivery.

PR #111, opened 2021-03-08
clean foler with move deliveries_ares_protocol_milestone_1.md to folder

xrdavies:
* clean foler with move deliveries_ares_protocol_milestone_1.md to deliveries and remove the prefix


alxs:
Thanks for noticing this!

PR #140, opened 2021-04-13
Curve amm miilestone 1

mikolajsobolewski:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [x] The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out for this milestone. 
- [x] This pull request is being made by the same account as the accepted application. 
- [x] In case of acceptance, the payment will be transferred to the initial BTC/ETH payment address. 
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).

PR #141, opened 2021-04-13
Curve amm miilestone 2

mikolajsobolewski:

- [x] In case of acceptance, the payment will be transferred to the initial BTC/ETH payment address. 
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).

### DIFF
--- a/deliveries/curve_amm-milestone_2.md
+++ b/deliveries/curve_amm-milestone_2.md
@@ -1,0 +1,27 @@
+# Milestone Delivery :mailbox:
+
+> ⚡ Only the GitHub account, which is responsible for the pull request of the accepted application is allowed to submit milestones. 
+> 
+> Don't remove any of the mandatory parts presented in bold letters or as headlines! Lines starting with `>`, such as this one, can be removed.
+
+**The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).**  
+
+* **PR Link:** https://github.com/w3f/Open-Grants-Program/pull/225
+* **Milestone Number:** #2
+
+> Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which will be used for testing. If you don't provide a commit hash, we will work off the default branch of your repository. Thus, if you plan on continuing work after delivery, we suggest you create a separate branch for either the delivery or your continuing work. 
+> 
+> If there is anything particular about any of the deliverables we or a future reader should know, use the respective `Notes` column.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1.a. | We will implement methods to work with asset pools: add liquidity.  | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/lib.rs#L406 |  | 
+| 1.b. | We will implement methods to work with asset pools: remove liquidity.  | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/lib.rs#L681 |  |
+| 2. | We will implement methods to exchange assets within pools. | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/lib.rs#L582 |  |
+| 3. | We will implement liquidity superfluidity: assets locked inside Curve liquidity pools may be further used in various lending protocols across the Polkadot ecosystem. |  | Any runtime that successfully integrated Equilibrium Curve Amm Pallet is able to generate LP tokens using underlying asset implementation. |
+| 4. | We will implement a mechanism to reward liquidity providers with LP tokens.  | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/lib.rs#L624 | LP tokens represent balances supplied to the protocol. Users earn interest from exchange fees through the LP token's exchange rate appreciation when LP tokens increases in value relative to underlying assets. |
+| 5. | The code will have complete unit-test coverage to ensure functionality and robustness. | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/tests.rs |  |
+| 6. | We will publish the code in the Equilibrium’s public GitHub repository. | https://github.com/equilibrium-eosdt/equilibrium-curve-amm |  |
+| 7. | We will provide detailed technical documentation describing pallet logic, storage, interfaces, subscriptions. | https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/milestone-2/pallets/equilibrium-curve-amm/src/lib.rs | See rustocs and comments throughout entire file. |
+
+


alxs:
Hey again @mikolajsobolewski. Looks good at first sight. I think there is just one deliverable that is not covered by the current delivery, which is an integration guide.

This is part of deliverable 0b both in M1 and M2, and besides I think it would really help other teams wanting to use this.

Could you add such a guide containing instructions on adding the pallet, a few code examples on how to use it, explaining the different features, generating LP tokens etc.? Ideally this would be written by a technical writer and other teams would be able to follow it to integrate Curve into their project.

Also is your LP tokens implementation different from Curve's original one? I'd appreciate it if you could point me to Curve's documentation on these. Same for liquidity superfluidity. Was this feature always supposed to be the same as creating LP tokens?

mikolajsobolewski:
Thank you for the comments, @alxs 

1. We will update documentation to include more details on how to add the pallet to one's substrate and also will provide some examples explaining features like setting up pools and generating LP tokens.

2. Our implementation of LP tokens is similar to Curve's original. LP tokens represent a share in the pool and their price rises as fees accrue when users exchange assets in the pool. They're very similar to Compound's C-tokens in this sense. For the liquidity superfluidity Curve in Ethereum has special-purpose pools where they support wrapped assets from other protocols such as Aave for example ( https://github.com/curvefi/curve-contract/tree/master/contracts/pools/aave ), so users who deposit liquidity into Aave, get a-tokens, which can in turn be deposited into special-purpose Curve pool - this is their way of handling superfluidity. In our case, since Polkadot ecosystem is not interconnected yet, and we don't know the details of XCMP implementation, we see superfluidity slightly differently: we (At least with equilibrium) will use LP tokens as a collateral or bailsman liquidity within our protocol, so users who deposit assets/liquidity into the Curve pools will have an ability to deposit their LP tokens into Equilibrium and borrow against them or earn rewards as insurers, so this is basically a reciprocal of what we see in Ethereum.

alxs:
Great, let me know when you're done and I'll be happy to review it.

Many thanks for the detailed explanation. I'm not sure I follow though. Aren't you talking about two different concepts here? I.e. it seems to me for Ethereum you're talking about "incoming" superfluidity i.e. repurposing liquidity that is locked in another protocol, while on Equilibrium you describe a version of the "outgoing" side of this i.e. leveraging liquidity locked in Curve pools. The deliverable also rather describes the latter.

I understand what you're saying though and why you'd deliver this. How is the "outgoing" side of this accomplished in the original Curve implementation? Can LP tokens also be used to borrow against etc. or how can liquidity locked in Curve pools be used elsewhere in Ethereum?

mikolajsobolewski:
Hi, @alxs

Definitely, you may look at the concept from two sides, using your terminology:

1. Outgoing: Users lock assets in some DeFi protocol, and later use tokens which represent their share of liquidity in that protocol in our protocol
2. Incoming: Users lock assets inside our protocol, use tokens which represent their share of liquidity in our protocol in some other DeFi protocol (e.g. setup a Curve pool). 

The outgoing side in Curve is accomplished via LP tokens and I believe they even considered a possibility to list those LP tokens in Lending protocols such as Aave or Compound. 

If we're talking explicitly about Equilibrium's use case, it is optimal for us to allow the use of LP tokens from Curve pools inside Equilibrium, since they're low volatility, rising price assets, which mitigates risks of borrowers and bailsman alike inside Equilibrium. 

The other side works as well, though. If some protocols on Polkadot decide to introduce share-tokens, such as Acala's L-DOTs, for example these L-DOTs may be later used to setup a Curve pool with some other wrapped-DOTs from other protocols.

alxs:
Sounds good. I was only wondering if this was really the intended deliverable, but as I said this makes sense and you certainly put enough effort into the project in general. Pending the discussed additional documentation, I'll be happy to accept the milestone. And kudos on the already created documentation, this would be the envy of most other pallets out there ;)

mikolajsobolewski:
Hello @alxs 
We've updated our docs and also created the detailed integration guide per your comments. https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/master/docs/GUIDE.md
Please check! Looking forward to your reply.

alxs:
Thanks, looks great. I'll look into it early next week. By the way, I left a comment regarding the invoice for the last milestone here https://github.com/w3f/Grant-Milestone-Delivery/pull/140#issuecomment-823559911. Just letting you know in case you missed it.

alxs:
Looks good. I only have two things to add:

- Are there API calls to retrieve pool information from the pallet? The guide seems to just assume that this would be handled on the client side, but in the case of e.g. an interface that provides access to Curve on different parachains this would need to be retrieved. I'm referring to the number of pools and the different `AssetIds` in them in specific, plus as much information on the assets as necessary. Would this make sense?
- It'd be good if the integration guide could be proofread since it currently contains quite a few grammar issues.

The issue with the invoice applies to this milestone too.

Veniamin:
1. We will correct our English in the guide.
2. Where do we submit invoices with the correct address? 
3. We will add description of internal API calls to the guide.

alxs:
Sounds good! You can submit the updated invoices the same way you submitted the current ones.

r0t0r-r0t0r:
@alxs, we've updated our pallet.

Here is the list of changes:
- The trait [`CurveAmm`](https://github.com/equilibrium-eosdt/equilibrium-curve-amm/blob/783172714d5fa25f5065736776ab8d3659de6726/pallets/equilibrium-curve-amm/src/lib.rs#L1486) was introduced. Now other pallets can use it to communicate with our pallet.
- Structure of the documentation was improved: client API documentation moved from integration guide; both integration guide and client API documentation are now available through links on the main [README.md](https://github.com/equilibrium-eosdt/equilibrium-curve-amm/tree/783172714d5fa25f5065736776ab8d3659de6726#using-the-pallet) file.
- Many grammar and punctuation errors were fixed.

The current rustoc documentation for `CurveAmm` trait is good enough. But let me know if you have any ideas for improving it.

r0t0r-r0t0r:
Hi @alxs!

Any updates on the second milestone? We are looking forward to your feedback.

r0t0r-r0t0r:
@alxs, you're right. It didn't work for me either. But I just fixed this. Please, pull the latest version and check again.

alxs:
Thanks, everything works now.

What would be the proper way of generating asset IDs for LP tokens, ideally in a way that is compatible across chains? And could you quickly walk me through how the appreciation in value of LP tokens whenever assets are exchanged is implemented?

It would also be great if you could add a couple more examples to the client API, specifically querying the chain for existing pools and asset IDs within + balances thereof, predicting slippage, and error handling for anything that could go wrong i.e. trying to deposit more than one's balance, exchanging more than the available liquidity etc.

r0t0r-r0t0r:
> What would be the proper way of generating asset IDs for LP tokens, ideally in a way that is compatible across chains?

We designed the pallet in a way that it entirely depends on the external implementation of the asset ID:

```rust
trait Assets<AssetId, Balance, AccountId> {
  fn create_asset() -> Result<AssetId, Error>;
  // other declarations are omitted
}
```
We assume that a party using a pallet has to define its own strategy on asset id generation. We tried to be as generic as possible, given we don’t know yet the specifics of the XCM protocol and cross-chain communications.

> And could you quickly walk me through how the appreciation in value of LP tokens whenever assets are exchanged is implemented?

Let’s say I’ve set up a pool with 10 renBTC and 10 polkaBTC, I get 20 LP tokens in return. The value of each LP token is currently 1 "BTC"

Further, a user comes in and buys 5 renBTC from the pool for 5 polkaBTC - 1% fee, that means user will receive only 4.95 renBTC after fees, so the pool will hold: 15 polkaBTC and 5.05 renBTC, while there are still 20 LP tokens out there. So the current effective price of 1 LP token is (15 + 5.05)/20 =  1.0025 "BTC" so the price rose from 1 to 1.0025.

> It would also be great if you could add a couple more examples to the client API, specifically querying the chain for existing pools and asset IDs within + balances thereof, predicting slippage, and error handling for anything that could go wrong i.e. trying to deposit more than one's balance, exchanging more than the available liquidity etc.

Please elaborate what you mean and what do you want to be added? We have all the checks and error handling implemented in the rust code. Providing more examples for a client library is out of scope of the milestone deliverables.

alxs:
Thanks for the thorough explanation.

Fair enough, in that case I'm happy to tell you that the milestone is a pass. You can find my evaluation notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/curve-amm_2_alxs.md).

I've submitted a PR to this branch to update the links in the delivery to the latest commit, as the line numbers were broken again. Feel free to review it and once it's merged I'll merge the delivery too. I'll already forward your invoice internally. I also forwarded the updated invoice for the first milestone a couple of days ago; you should receive payment for both of them soon.

RouvenP:
hi @mikolajsobolewski and @r0t0r-r0t0r we transferred the payment for milestone 2.

alxs:
@r0t0r-r0t0r @mikolajsobolewski @Veniamin I just noticed that you mentioned in the application that you will deploy this on Equilibrium. Just wondering about the timeline for this?

Veniamin:
We will add to genshiro chain approximately at the end of July, everything was agreed in the builders program with Santiago

PR #238, opened 2021-07-23
Minor updates to the README

alxs:
Only significant change is updating the reward for external evaluations to "up to 500 USD in KSM".


mmagician (approved):
looks good

PR #242, opened 2021-07-27
Evaluation of Standard Protocol, round 1

mmagician:


hskang9:
@mmagician the delivery has updated to the evaluation.

PR #243, opened 2021-08-01
Webb extended mixer delivery

drewstone:

- [x] The [invoice form :pencil:](https://forms.gle/vgQEZzvvFUjCyBiW6) has been filled out for this milestone. 
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md). 

Link to the application PR: https://github.com/w3f/Open-Grants-Program/pull/422

### DIFF
--- a/deliveries/substrate-molochdao-milestone-2.md
+++ b/deliveries/substrate-molochdao-milestone-2.md
@@ -1,0 +1,16 @@
+# Milestone Delivery :mailbox:
+
+> Only the GitHub account, which is responsible for the pull request of the accepted application is allowed to submit milestones. Don't remove any of the mandatory parts presented in bold letters or as headlines!
+
+**The [invoice form :pencil:](https://docs.google.com/forms/d/e/1FAIpQLSdSqj2vYjvpiIytkjcc40Pwl0Eg76WGUAq5L9e8eFuuOegmLw/viewform) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).**  
+
+* **Application Document:** https://github.com/w3f/Open-Grants-Program/blob/master/applications/dora-factory-molochdao-v1-v2.md
+* **Milestone Number:** 2
+
+Please provide a list of all deliverables of the milestone extracted from the initial application and a link to the deliverable itself. Ideally all links inside the below table should include a commit hash, which should be used for testing.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| 1. | MolochDAO v1 Pallet on Substrate| https://github.com/DoraFactory/Substrate-Moloch-V2 | This pallet impplements MolochDAO v2 on Substrate. Key functions include: summon a MolochDAO (set up period duration, voting and grace period periods, etc.), create a proposal, request shares and loots, submit a proposal, sponsor a proposal, vote for a proposal, process a proposal, guild bank, guild kick, ragequit, abort proposal, and update delegate. | 
+| 2.  | Test Guide | https://github.com/DoraFactory/Substrate-Moloch-V2/blob/main/doc/test-guide.md | Follow test guide to test all functions of MolochDAO v1 | 
+| 3.  | Unit Test | https://github.com/DoraFactory/Substrate-Moloch-V2/blob/main/pallets/moloch-v2/src/tests.rs | Use unit tests in test.rs | 
--- a/evaluations/molochDAO_2_semuelle.md
+++ b/evaluations/molochDAO_2_semuelle.md
@@ -9,8 +9,8 @@
 | Number | Deliverable | Accepted | Link | Evaluation Notes |
 | ------ | ----------- | :------: | ---- |----------------- |
 | 0a. | License | <ul><li>[x] </li></ul> | [license](https://github.com/DoraFactory/Substrate-Moloch-V2/blob/3e000a08679609029276767cdb013edcfe6db12d/LICENSE) | MIT |
-| 0b. | Testing Guide | <ul><li>[x] </li></ul> | [gh](https://github.com/DoraFactory/Substrate-Moloch-V2/blob/3e000a08679609029276767cdb013edcfe6db12d/doc/test-guide.md) | tarpaulin reports 93% code coverage |
-| 1. | MolochDAO v2 Substrate Pallet | <ul><li>[x] </li></ul> | link | see [General Notes](#general-notes) |
+| 0b. | Testing Guide | <ul><li>[x] </li></ul> | [gh](https://github.com/DoraFactory/Substrate-Moloch-V2/blob/ced0a467e80d1f11f74f44dc5f922e30f93872ca/doc/test-guide.md) | tarpaulin reports 93% code coverage |
+| 1. | MolochDAO v2 Substrate Pallet | <ul><li>[x] </li></ul> | [gh](https://github.com/DoraFactory/Substrate-Moloch-V2/blob/ced0a467e80d1f11f74f44dc5f922e30f93872ca/pallets/moloch-v2/src/lib.rs) | see [General Notes](#general-notes) |
 | 2. | Test | <ul><li>[x] </li></ul> | [gh](https://github.com/DoraFactory/Substrate-Moloch-V2/blob/3e000a08679609029276767cdb013edcfe6db12d/pallets/moloch-v2/src/tests.rs) | — |
 
 


Noc2:

deliveries/Webb_Mixer_Extended_milestone_1.md
```diff
@@ -0,0 +1,24 @@
+# Milestone Delivery :mailbox:
+
+Payment in DOTS to: 16aN6hVU7WtMD7pw2kjXNwnApQB9tNxcdsknTSyFDps6YAiA
+
```
alxs, line 4:
```suggestion
```

deliveries/Webb_Mixer_Extended_milestone_1.md
```diff
@@ -0,0 +1,24 @@
...
+**The [invoice form :pencil:](https://docs.google.com/forms/d/e/1FAIpQLSfmNYaoCgrxyhzgoKQ0ynQvnNRoTmgApz9NrMp-hd8mhIiO0A/viewform) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).**  
+
+* **Application Document:** https://github.com/w3f/Open-Grants-Program/pull/422
```
alxs, line 7:
```suggestion
* **Application Document:** https://github.com/w3f/Grants-Program/blob/master/applications/MIXERv2.md
```

deliveries/Webb_Mixer_Extended_milestone_1.md
```diff
@@ -0,0 +1,24 @@
...
+| 0d. | Article/Tutorial | https://www.notion.so/hicommonwealth/Blog-post-268628ee32ac4f21b9b8531385b00458 | 
+| 1. | Substrate module: Multi-Asset support for mixer | https://github.com/webb-tools/anon/tree/master/pallets |
+| 1a. | Substrate module: ORML support | https://github.com/webb-tools/anon/tree/master/pallets/mixer |
+| 1b. | Substrate module: Pallet asset support | https://github.com/webb-tools/anon/tree/master/pallets/assets-mixer |
+| 2. | Substrate module: New ZKP backend support | https://github.com/webb-tools/anon/blob/master/pallets/merkle/src/utils/setup.rs |
+| 2a. | Substrate module: New ZKP backend support | https://github.com/webb-tools/anon/blob/master/pallets/merkle/src/utils/setup.rs |
+| 2b. | Substrate module: ZKP backend support | https://github.com/webb-tools/anon/blob/master/pallets/merkle/src/utils/setup.rs |
+| 2c. | Substrate module: New ZKP hash function support | https://github.com/webb-tools/anon/blob/master/pallets/merkle/src/utils/setup.rs |
+| 3. | Zero-knowledge circuits | https://github.com/webb-tools/arkworks-gadgets |
```
alxs, line 23: Do I understand correctly that this is Apache 2.0, and you also include the MIT one since you're using MIT-licensed code? Could you state this more explicitly or rename the first file to just `LICENSE`?

alxs (commented):
This looks very good, I'll get back to you early next week with some feedback. Regarding a follow-up grant, I've pinged a few of our researchers (seems you already met Al and Fatemeh at EthCC!) and DKG in some form is what was generally received with most enthusiasm. You can reach me on Element under `@aleixo:web3.foundation` if you'd like to chat.

deliveries/Webb_Mixer_Extended_milestone_1.md
drewstone: Yep, all done.

alxs:
Btw, it will be end of the week :)

alxs:
Still getting my teeth into this, but I have a few initial comments:

- `cargo clippy` reports a number of errors and warnings for both *anon* and *arkworks-gadgets* - wanna take care of those?
- The tests understandably timed out for me, could you add a brief section to the READMEs with testing instructions?
- Do you intend to publish the new pallet(s) and could you update the ones you already did?
- Could you provide some usage instructions for https://github.com/webb-tools/arkworks-gadgets?
- I feel like there might be some useful information in the article that is not included in the README or the documentation - not 100% sure yet, but maybe you want to check. The repo + documentation would ideally be self-contained, naturally.

drewstone:
@alxs I will add more docs to README. You should also make sure to run the tests in `--release` mode. The cryptographic operations are far slower in `--debug` if that's how you are testing.

drewstone:
@alxs I updated the readme with test script, you should only need to run `./scripts/test.sh`.

drewstone:
Just fixed the clippy errors too @alxs. Let me know what's remaining.

drewstone:
Any update @alxs @Noc2 /

alxs:
On it @drewstone. Will follow up with an update this week. Apologies for the delay.

alxs:
@drewstone I'll leave a full evaluation tomorrow evening. Just wondering, how far are you with milestone 2?

drewstone:
Fairly far along. We’ve modularized the design of the mixers to optimise for parameter storage in this repo: https://github.com/webb-tools/darkwebb-substrate. 

We’ll be building the UI against that and are decently through the integration for the phase 2 of the grant. Just need to hook it up but most of the toolchain exists.

alxs:
Hey @drewstone, thank you very much for your patience. You can find my evaluation notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/Webb-Mixer-Extended_1_alxs.md). I'll let you know whenever I update them. Happy to discuss them via chat but please leave a reply eventually here on GitHub so the discussion is available to everyone to see.

drewstone:
Thanks for the deeper feedback @alxs.

(0a).  I fixed the license, it is now Apache 2.0
(0b). I've added some high-level pallet discussion too: https://github.com/webb-tools/anon/tree/master/pallets. It is short but overall the main docs cover how the system's modular pieces fit together.
(0c). To be fairly honest, we don't develop in any other mode other than release for doing ZKP work. The tests are likely to fail and time out if you do not use release mode. This is not something I've investigated as `debug` mode is tremendously slower w/ zero-knowledge computations anyway, so `release` mode is our default. I would not suggest looking at it in any other way. The backends may be restricted to only a few options, this is because the circuits in `arkworks-gadgets` actually generalize over any curve. So we're confident we can plug and play since its the same circuit underneath the hood (for arkworks dependant mixers).

3. We've actually continued to update these gadgets to reduce code duplication. We are working quite fast on a variety of parts of our protocol and given that this review has taken a long time have continued to update things to improve the code for new repos and not `anon`. Recently we abstracted a lot of the code to use macros to improve the duplication. https://github.com/webb-tools/arkworks-gadgets/pull/43.

The bridge pallet / gadgets should be avoided for now as they are also not part of this grant or any part of the proposal. That was code we added to build out extended mechanisms using this architecture. It is in progress and is built off of the mixer work so you may see duplication still there, though please do not regard it as part of this grant. To answer more though, we use Groth16 style zkSNARKs which are circuit-specific SNARKs. This means that we cannot reuse a gadget between two different systems, as those systems may have different constraints and complexities under the hood. Our `arkworks-gadgets` is designed so we can combine modular pieces together in aggregate circuits (for mixers or bridges) but they do need to be defined and exist separately since they have different computations going on under the hood.

4. I will check on this docker issue.

drewstone:
Hey @alxs I just removed an old stale Dockerfile from the root directory but the one in `docker/Dockerfile` builds fine for me. Can you try that?

alxs:
Thanks @drewstone for the thorough reply and a great chat last Friday. Below are a few follow-up comments and leftover issues (I won't update my notes for now as it seems more convenient to continue the conversation here, you can forget about them):

[anon]
- The pallet documentation the README points to is outdated. I'd be happy to see the mixers + merkle pallets published/updated as previously, but feel free to instead amend this section to point to the README inside the pallets folder + provide instructions to build the rustdocs.
- The Docker instructions still don't work for me. As I mentioned there's no need to publish prebuilt images if you prefer to go with the Dockerfile, but the instructions in the README should work.
- As discussed on Friday, could you look anyway into why those tests would fail with an underflow in debug mode? I quickly looked it up and under & overflow checks are indeed disabled in release mode so it's most certainly also happening there, it's just not caught.
- Some documentation like the newest version of the README in `arkworks-gadgets` would be fantastic also for this repo, but if `darkwebb` completely supersedes this, agree, no need to put in unnecessary effort. So this is rather a recommendation for future deliveries and in general. Nevertheless, note that while I understand `darkwebb` is still the same project at its core and you've simply moved on to using an improved design, we still care about `anon` being functional, correct and accessible, ideally maintained for a while, since it is after all the main deliverable of a grant(s) we've invested close to $100k into.

[arkworks-gadgets]
- Kudos on the additional documentation, it looks great and should be of huge help to anyone interested in using your work.
- Regarding the specific gadgets/circuits, we also already briefly touched on this but the application mentions a "non-zero circuit and bounded number circuit", which I think aren't currently implemented, but trivial and hence superfluous. Could you confirm?
- "Support for fees inside these circuits" is also mentioned, how were you picturing this? Could you add some instructions on using this for the general case? The only place I can find anything of the sort is in the mixer, which [as you point out](https://github.com/webb-tools/arkworks-gadgets#public-input-structure) has a fee variable in the public inputs. However as far as I could see this input seems to only be constrained to be a valid point on the curve (I assume this is what [this](https://github.com/webb-tools/arkworks-gadgets/blob/master/src/arbitrary/mixer_data/constraints.rs#L39) is?), as all the other public inputs. I could absolutely be missing something here but otherwise, my question would be what the point of the public inputs is then, and in general why the need to match Tornado Cash's design/inputs?
- I honestly feel like testing should be significantly expanded, e.g. most of the gadgets and circuits are only tested using one curve or hash function, which could easily miss faulty implementations, copy-paste errors etc for other parameters. But more importantly, I couldn't find any test that tested more than initialisation and a dummy case.
- Seems [some of the gadgets](https://github.com/webb-tools/arkworks-gadgets/blob/master/src/merkle_tree/mod.rs#L105) were inspired by [existing ones](https://github.com/arkworks-rs/ivls/blob/master/src/building_blocks/mt/merkle_sparse_tree/mod.rs#L313). Kind of obvious especially since this is from arkworks itself, but I'd ask you to please still attribute the original authors. You can add a notice such as [this](https://pastebin.com/raw/YjKk07L9) to the beginning of the relevant files. 

On an unrelated note, for when you get around to setting up a server to build&test these projects on (which again I'd highly recommend) - [cargo-remote](https://github.com/sgeisler/cargo-remote) is a nifty little tool.

drewstone:
Will address all these items soon, for now here are some more comments @alxs .

1. I just ran `cargo test` on arkworks-gadgets locally and didn't see any failed tests even in debug mode. Not sure what to make of this except that maybe our machines are different.
```
test result: ok. 56 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1174.58s
```

2. The non-zero and bounded number gadgets are trival and unimplemented because we don't use them in these mixer gadgets anyway.
3. The fees, refund, etc. are just multiplied by themselves to add constraints to the circuit in order to generate unique proofs that are bound to these values. Because these mechanisms are built with transaction relayers who set these fees, we must ensure that these values can't be tampered with. You can imagine these values being set on-chain and provide directly from a blockchain's storage (smart contract or runtime pallet)
4. Will definitely add copyrights for  origin authored files.

drewstone:
Yea, it's not friendly to publish to cargo docs since the repo uses git dependencies (since substrate hasn't been stable w/ crate releases). I will amend the docs part and add instructions for compiling them locally.

alxs:
Sounds good! So the failed tests were in `anon`. Regarding fees, I think I can see what you mean although it'd seem easier and more efficient to just provide a signature of these values concatenated with the proof? In any case, I wouldn't say this really constitutes "support for fees inside these circuits" which I'd understand as implementing a fee payout within the circuit (which only really makes sense for a circuit spending a leaf to create a new one?), whereas the current logic basically just commits to a number and is only implemented in one circuit. I'm not sure how you were picturing this when you applied?

Besides, I see these public inputs are actually used when setting up the circuit which doesn't make much sense to me. On closer examination, the functionality the mixer provides doesn't seem complete - it seems to only be possible to add a leaf and a new circuit is instantiated every time this is done. I also couldn't find any usage of it in `anon` and it appears closer to a dummy example than an actual implementation of a mixer. Am I wrong, or how is it meant to be used?

And a few miscellaneous comments:

- I may be really missing something here, but why is the mixer referred to as 'a circuit' in the first place? Doesn't a mixer consist of usually two circuits (or more but bundled up into two steps 'proving that a new leaf is correct' and 'proving that I can spend this leaf'), and then all the logic around that?
- What's the purpose of [these](https://github.com/webb-tools/arkworks-gadgets/blob/master/src/setup/mixer.rs#L356)?
- Is there any actual separation between Provers and Verifiers as [described](https://github.com/webb-tools/arkworks-gadgets#provers) in the readme?
- Is there any use in the [parameter generation](https://github.com/webb-tools/arkworks-gadgets#parameter-generation) being explained in the readme? Seeing as these are all already generated and this information is included in the individual files containing the parameters, I think this can be removed.

drewstone:
>  I think I can see what you mean although it'd seem easier and more efficient to just provide a signature of these values concatenated with the proof?
This would not work because these transactions are relayed by untrusted tx relayers. Who submits the signature? What key is it signed under and how is that validated? Seems like this exposes privacy if done in any meaningful way.

This is exactly how Tornado Cash handles fees and the design does make a lot of sense. The proof is bound to the fee and thus you can execute payout and handle the arithmetic on the smart contract or blockchain that verifies such a proof with these public inputs. The point is that the transaction relayers can't tamper with the fee after the user has selected it, they can only _not_ submit it for the user.

> Besides, I see these public inputs are actually used when setting up the circuit which doesn't make much sense to me
Because these public inputs create extra constraints and because we use Groth16, which is a circuit specific zkSNARK, these values are directly relevant towards circuit setup and construction, especially when creating the proving/verifying keys. These keys are all that is needed to make successive proofs against this type of circuit.

> it seems to only be possible to add a leaf and a new circuit is instantiated every time this is done. I also couldn't find any usage of it in anon and it appears closer to a dummy example than an actual implementation of a mixer. Am I wrong, or how is it meant to be used?
Yes, this is a misunderstanding. The APIs for the utilizing these circuits rests on how the proving/verifying keys are used and generated. You need to instantiate the `Circuit` object every time you make a proof. If you want to make a proof that verifies successfully, then you must use the proving key that corresponds to the verifying key that was created when they both were created. We create these once and stick the verifying key on-chain. Thus, if you make proofs you must instantiate a circuit and use the corresponding proving key when generating the proof.

> I may be really missing something here, but why is the mixer referred to as 'a circuit' in the first place? Doesn't a mixer consist of usually two circuits (or more but bundled up into two steps 'proving that a new leaf is correct' and 'proving that I can spend this leaf'), and then all the logic around that?
This is just a semantical question and I don't see the relevance tbh. We have called intermediate components gadgets and the end result a circuit. The mixer circuit executes gadgets that verify a leaf was created properly with a secret and nullifier and then we have a gadget that verifies the leaf is in the tree. We don't need to prove that a new leaf is correct on deposit. If it isn't correct when inserted into the tree, then it's impossible to successfully prove that the leaf was created properly on withdrawal by the properties of collision-resistant hash functions.

> What's the purpose of these?
This is to mock smart contract / pallet calls to verify. They are not needed and we can remove if this is really a sticking point.

> Is there any actual separation between Provers and Verifiers as described in the readme?
Yes there is, this is directly related to implementations of this system in part 2 of the grant. We embed provers in the browser, in WASM, so that we can generate proofs and submit them on-chain. We embed the verifiers (generically they are Groth16 verify functions) inside the pallets.

> Is there any use in the parameter generation being explained in the readme? Seeing as these are all already generated and this information is included in the individual files containing the parameters, I think this can be removed.
None other than making it clear from the README that these are sensitive parameters and specific to each instantiation of a circuit and mixer application. We can remove this from the readme.

alxs:
@drewstone thanks for facilitating the discussion, let me get back to you on the longer bits tomorrow - just a reminder to address those updates and a few leftover items you wanted to take care of from previous comments.

drewstone:
@alxs we fixed the broken tests in `anon` too and I removed the parameter gen from the README.

Let me know any next steps.

RouvenP:
hi @drewstone we transferred the payment today.

PR #165, opened 2021-05-05
Spartan PoC Consensus: Milestone 1

rg3l3dr:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [x] The [invoice form :pencil:](https://forms.gle/8Wx7nxtq8fKrsuEz8) has been filled out for this milestone. 
- [x] This pull request is being made by the same account as the accepted application. 
- [x] In case of acceptance, the payment will be transferred to the initial BTC/ETH payment address. 
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md).

PR #245, opened 2021-08-02
Spartan PoC Consensus: Milestone 3

rg3l3dr:

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/357


Noc2:

Noc2 (changes requested):
Thanks again for the delivery and sorry for the late reply. In general it looks good. You can find my notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/spartan-poc-consensus-module_3_Noc2.md). I have a few questions:

- It would be nice if you could create a separate repo, which only contains your chain/pallets instead of a fork of substrate (even if you still mostly develop inside the fork). This way it will be easier for others to understand and potentially leverage your work. Also it’s quite difficult for example to search for specific functions if you code is in one big repo with the complete substrate implementation. Furthermore rust-analyse goes nuts everytime I open the substrate repo in vscode ;-)
- As far as I can see there is no “inline documentation” regarding the attacks/security analysis. It doesn’t need to be inline documentation as specified in the contract, but it still would be useful if this information is publicly available inside the repo, not only as a notion document. 
- Your work regarding the equivocation attack seems to be unfinished, [see](https://github.com/subspace/substrate/blob/b84b84f47326955978a41680e7783d6e829b5d73/bin/node-template-spartan/runtime/src/lib.rs#L487)
- I had to use cargo +nightly test in order to test [spartan](https://github.com/subspace/substrate/tree/poc/frame/spartan)
- Reading the notion document, it seems to me that you still want to look into Safety and Liveness. Is this still the case? Otherwise we need to amend the original contract. 
- Also how do you come up with the  47% in the medium article.

nazar-pc:
> It would be nice if you could create a separate repo

This point was already raised before in the [discussion about the first milestone](https://github.com/w3f/Grant-Milestone-Delivery/pull/165). We've described the rationale there, which I hope explains why we have chosen this approach. We do consider the approach you have suggested and will probably incorporate it in future updates (we do not plan to stop after this milestone :wink: ). In the meantime GitHub's diff should give an overview of what we have changed/added: https://github.com/paritytech/substrate/compare/master...subspace:w3f-spartan-ms-3

> Your work regarding the equivocation attack seems to be unfinished

It is finished. The equivocation mechanism the way it functions is similar to BABE's: it will ban farmer permanently once equivocation proof is included into the block. However, there are still a few potential slots where farmer can continue to produce multiple solutions for the same block until the block is found by other nodes on the network. TODO describes a minor potential improvement that we may or may not do, just a note to think about it in the future, not something that needs to be done immediately as it already functions as expected.

> I had to use cargo +nightly test in order to test

Good point, I have updated those and the rest of commands in readme to include `+nightly` (we typically set nightly by default for the whole project, so we don't type `+nightly` all the time, that is why it was missing there).

---

My colleagues will reply to other points a bit later. Thanks for quick review!

ssrivatsan97:
> * Reading the notion document, it seems to me that you still want to look into Safety and Liveness. Is this still the case? Otherwise we need to amend the original contract.

In the current update, we have implemented countermeasures against all the major attacks and analyzed their effectiveness. There are a couple of attacks that we are still looking into. Particularly, there is a long-range (aka history rewriting) attack which is common across all PoS and PoC designs, and a short-range space-time trade-off attack. We are currently working on a proof-of-time solution which will make the long-range attack infeasible and the space-time trade-off economically irrational. The solution and its effectiveness will be outlined in another post as soon as we are done with it. 
By dealing with these attacks which are well-known in PoS and PoC literature, we would justify our assumption that Subspace is a PoC extension of Bitcoin and Ouroboros Praos. Moreover, our economic model shows that replacing storage by computation is irrational. This will ensure that Subspace inherits the safety and liveness properties of Ouroboros Praos, as long as no single rational adversary holds more than 47% of the total network's storage.

> * Also how do you come up with the  47% in the medium article.

We use the mathematical analysis from https://arxiv.org/abs/1910.02218 to get to this number. The linked paper has a thorough security analysis of the described solution against the simulation attack. The paper derives a formula for the security threshold as a function of the number of blocks that share a randomness, from which we get 47% for 256 blocks. The paper is linked in the Notion doc along with a plot showing the results of this formula.

I will also add the Notion doc as a Markdown file to our github repo for convenient access.

Noc2:
Thanks for the detailed reply. @nazar-pc I would still argue that it’s useful to publish the code in a separate repo, but it’s not mandatory for the acceptance of this contract. @ssrivatsan97  Let me know once and where you integrated the notion into the github repo.

ssrivatsan97:
@Noc2 the notion writeup is now at https://github.com/subspace/substrate/blob/w3f-spartan-ms-3/frame/spartan/ms3-security.md

Noc2:
Thanks for the update. I’m happy to tell you that the milestone is a pass and I will forward your invoice internally.

Noc2:
@rg3l3dr I only found the invoice for milestone 2. Could you send us the invoice for milestone 3? https://docs.google.com/forms/d/e/1FAIpQLSfmNYaoCgrxyhzgoKQ0ynQvnNRoTmgApz9NrMp-hd8mhIiO0A/viewform

rg3l3dr:
@Noc2 Thanks for your feedback and quick review! 

I did submit an invoice about one week ago, so that's strange that it does not show up. Perhaps I used the wrong naming convention (Subspace Labs instead of spartan poc consensus module). In any event, I just resubmitted everything so you should see it now.

Noc2:
Sorry, our mistake. Apparently we changed the upload folder. Anyway, we received the invoice. Thanks a lot!

CourtneyZorio:
Good morning (for me anyway!) @Noc2 ,

I'm following up on the invoice that was submitted by my colleague on August 10th for the completion of Milestone 3. We have not received that payment yet. Do you have an update for me or any questions I can answer? 

Thanks!

mmagician:
@CourtneyZorio Apologies for this. I will heck the status of your payment now and will get back to you soon with more info.

mmagician:
It seems that we missed this payment. I have got the confirmation that it is booked now and will be paid out by the end of the week. Sorry for the inconvenience caused!

CourtneyZorio:
@mmagician Thank you for the update, appreciate you looking into this.

PR #299, opened 2021-11-02
Actions revamp

alxs:
- Leave badge comment on first delivery
- Retry sheet update until it succeeds
- Check author in separate workflow
- Archive local testing files to be created from readme


Noc2 (approved):
I haven’t tested it, but I'm happy to test it in production.

PR #368, opened 2022-02-15
Blackprint visual programming for Polkadot.js - M1

StefansArya:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [x] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [x] This pull request is being made by the same account as the accepted application.
- [x] In case of acceptance, the payment will be transferred to the BTC/ETH payment address provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/Grants-Program/blob/master/docs/milestone-deliverables-guidelines.md).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/774

PR #377, opened 2022-02-28
Blackprint milestone 1 eval

cruikshankss:
Application: https://github.com/w3f/Grants-Program/pull/774
Milestone 1: https://github.com/w3f/Grant-Milestone-Delivery/pull/368

Copied from Conclusion:

This seems like really excellent, high-quality codebase. With such high-quality comments and documentation, I expect it will be possible for Blackprint to build new features and maintain the code. It seems very accessible to devs who wish to join an open-source project.


Noc2 (approved):
Going forward feel free to merge them yourself

StefansArya:
Thank you for the great evaluation!

> You could potentially look into using `yarn` instead of `npm` in the future for security purposes.

Thanks! actually you can just replace `npm` with `yarn` when installing dependencies. Personally I was using `pnpm` and already think to replace `gulp` and `browser-sync` with another package to reduce the vulnerability stats. But I just use it for development environment only, and not something to be distributed/bundled for production. I'm actually hope the maintainer of these package updated their dependency version thought.

> For absolute beginners, you could potentially make it more clear the editor is live here and I don't have to clone and run a local instance, because I was confused at first.

Ohh, sorry. I actually already put some links that may lead to the online editor on the [nodes-polkadot.js](https://github.com/Blackprint/nodes-polkadot.js)'s repository (and also below the "About" description on the repository).

Maybe you're missed that one 😅, but maybe it was my fault too because I'm afraid if I put the links on the grant proposal or delivery it may be inconsistent in the future if I have changed the editor's URL (especially if I decided to create it's own website domain).

> Also, you could potentially explain what a "Dummy Key" is and how it's used, and what the word "Provider" means.

Yep, I already planned it for Milestone 3 (that may have better docs/guide for explaining things). In the future, when you hovering through the port or node it will show a tooltip that have a description about what the port is and what the purpose of the node right on the editor.

But it's not easy to explain everything (especially when the module have so many nodes and input/output ports) and I may need some time to maintain and make the tooltip/docs works efficiently. The docs may also being bundled and distributed as a separated file so it can be loaded on the Editor and not in the production. But because it's still an unfinished feature, please expect something may be little different 😅.

PR #376, opened 2022-02-28
New BLS12 hash function M1

dishport:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [ ] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [x] This pull request is being made by the same account as the accepted application.
- [ ] In case of acceptance, the payment will be transferred to the BTC/ETH payment address provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/Grants-Program/blob/master/docs/milestone-deliverables-guidelines.md).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/966

PR #398, opened 2022-03-18
Evaluation of Pull request #376

drskalman:
Evaluation of [Milestone 1](https://github.com/w3f/Grant-Milestone-Delivery/pull/376) of https://github.com/w3f/Grants-Program/blob/master/applications/new_bls12_hash_function.md

### DIFF
--- a/evaluations/new_bls12_hash_function_1_drskalman.md
+++ b/evaluations/new_bls12_hash_function_1_drskalman.md
@@ -1,0 +1,47 @@
+# Evaluation
+
+- **Status:** In Progress
+- **Application Document:** https://github.com/w3f/Grants-Program/blob/master/applications/new_bls12_hash_function.md
+- **Milestone:** 1
+- **Kusama Identity:** N/A
+- **Previously successfully merged evaluation:** N/A
+
+| -----: | ----------- | ------------- |
+| Number | Deliverable | Accepted | Link | Evaluation Notes |
+| ------ | ----------- | -------- | ---- |----------------- |
+| 0a. | License | <ul><li>[x] </li></ul> | [GPL 3](https://github.com/w3f-community/Indifferentiable-hashing-to-ordinary-elliptic-curves-of-j-0-with-the-cost-of-one-exponentiation/blob/main/LICENSE)|
+| 0b. | Documentation |  <ul><li> [ ] </li></ul> | In line | In line comments are very sparse. Ideally every block of code should refer to the part/lemma/theorem from the paper that it is implementing| 
+| 0c. | Testing Guide | <ul><li> [ ] </li></ul> | | There is no unit test or doc test, need doc test for constant test vectors and at unit test to demonstrate covering almost the whole curve for a small field (see comment) | 
+| 0e. | Article | <ul><li> [ ] </li></ul>|  [Section 1.1](https://eprint.iacr.org/2021/1082) | I would give more weight to the cited implantation "We have provided a reference implementation ..." instead of "Incidentally". Perhaps also acknowledging W3F Grant in the acknowledgement section   |
+| 1 |  Implementation | <ul><li> [ ] </li></ul> | [Implementation](https://github.com/w3f-community/dishport/Indifferentiable-hashing-to-ordinary-elliptic-curves-of-j-0-with-the-cost-of-one-exponentiation) | Please provide a a unified (one) implementation of the hash function which depends on the bls12 parameter |
+
+## General Notes
+
+Thank you very much for submitting the deliverables:
+
+There are few points that we would like to see in sage implementation, in order to make it most beneficial for other researchers and implementors.
+
+1. Implementation
+
+Currently there are two (almost identical implementation) for hashing to bls12-381 and bls12-377. It would be most beneficial if you could merge them and implement the algorithm for hashing to bls curves as a single class or at least single set of functions which only depend on the defining bls parameter(s), the parameter then minimally changes the behaviour of the function instead of having two copy of each functions.
+
+0b. Documentation
+
+It would be helpful if you could add more comment to the code so it is clear that which lemma/case/equation in the paper, each step of implementation is related to.
+
+0c. Testing Guide
+
+We would like that if you could follow the Sagemath doctesting convention and add examples with expected output to all of your functions:
+
+ https://doc.sagemath.org/html/en/developer/doctesting.html#chapter-doctesting
+
+This could for example be output of pseudorandomly generated but hard-coded input which can later plays as reference for future developers to easily port the code to other languages and frameworks.
+
+It is very helpful that your code is testing that a random hashed point is actually resides on the target curve but if we could have some constant result in the test suite which similarly confirming the resulting points reside on the curve included in the doctests that would helpful.
+
+It is also helpful to add a test with constant value testing that result of phi is actually on the expected 3 fold. 
+
+Similarly, Having the defining equations of T from previous test, you could symbolically (perhaps as a scheme) test that T/F is an elliptic curve.
+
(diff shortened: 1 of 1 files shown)


alxs:
Thanks a lot @drskalman, this looks great. I would leave the PR open for now and you can keep editing it until you're happy with all deliverables.

dishport:
> drskalman

Thank you very much for your evaluation. In fact, my implementation covers not only the curves BLS12-381 and BLS12-377, rather than the cases 
1) q = 10 (mod 27) and b is not a cubic root in the field Fq.
2) q = 7 (mod 9) and b is a cubic root in the field Fq. In this case, it is sufficient to take b = 1, because b is always a square root in Fq by an assumption of my article.

I will add detailed comments in the Sage code on these two cases. Of course, I can handle them in the same Sage file in such a way that one can change q and b satisfying the conditions above. However, if you want a fast constant-time implementation, then in any case you need to handle independently all the reminders of q modulo 27 and mutually exclusive cases concerning the cubic residuosity of b. 

In other words, "unified (one) implementation" is not possible even in theory. On the other hand, in the real-world cryptography only the curves BLS12-381, BLS12-377 are applied, as far as I know. Therefore, an implementation of the cases 1), 2) should be sufficient for W3F purposes. To be honest, at the moment, I wouldn't want to do the boring handling of other cases, because it will be useless in practice. Instead, it would be much more important to rewrite the cases 1), 2) in Rust. 

What do you think ?

cruikshankss:
Good day @dishport, my name is Ashley and I work here at W3F. Today, I have replaced @alxs as the liaison between you and @drskalman. 

I reviewed your sage implementation I would like to respond to this point you raised: 

> In the implementation I stick to the notation of my article. I don't understand how to refer more precisely and how to comment in more detail.

If I'm not mistaken, I believe @drskalman may be asking you to provide details and comments to make your research more accessible to those without your level of knowledge. For example, would it be possible for you to add comments to your sage implementation to explain your research to me? I have only a background in Calculus, Discrete Mathematics, Computer Science, web development, and Rust. Is it possible for you to annotate your code to provide explanations at my knowledge level? If so, great! If not, I would appreciate to understand why not. 

The reasoning behind this comment is thus: we aim to steward your project to be as accessible as possible to the Substrate community, and so perhaps the closer your documentation approaches an ELI5, the more value others can derive from your research.

I am also curious about the point you made regarding testing (I believe):

>To be honest, at the moment, I wouldn't want to do the boring handling of other cases, because it will be useless in practice.

I am curious, why would more detailed unit testing of edge cases be useless in practice? My apologies that I do not natively understand, and I very much appreciate your time and effort in explaining this to me. I know your job is to research and not necessarily to teach, but documentation and testing guides are part of the contract for the grant, which is why I humbly request further information.

Kind regards,
Ashley

dishport:
Good day @cruikshankss.

> Would it be possible for you to add comments to your sage implementation to explain your research to me? I have only a background in Calculus, Discrete Mathematics, Computer Science, web development, and Rust. Is it possible for you to annotate your code to provide explanations at my knowledge level? If so, great! If not, I would appreciate to understand why not.

I cannot explain my research in the code more clearly than in my article. To fully understand the new hash function, you need to know at least the basics of algebraic geometry. Otherwise, look only at Section 2 of the article. 

> I am also curious about the point you made regarding testing (I believe). I am curious, why would more detailed unit testing of edge cases be useless in practice? My apologies that I do not natively understand, and I very much appreciate your time and effort in explaining this to me. 

It doesn't concern testing. I wrote the Sage code suitable only for the most important cases 1), 2) mentioned above. I wouldn't want to do the boring handling of other cases, because they don't correspond to real-world elliptic curves. 

As regards testing, I cannot test as professionally as it is said in
https://doc.sagemath.org/html/en/developer/doctesting.html#chapter-doctesting, because, frankly, this document is difficult for me. But I can only give naive tests, the way I see it.

Kind regards, Dimitri.

cruikshankss:
Hello @dishport and thank you very much for the explanations regarding documentation and testing. I find that very helpful.

@drskalman, please let me know if there's anything I can do to help your evaluation.

drskalman:
Thank you @dishport for going through my points.

> Of course, I can handle these two cases in the same Sage file in such a way that one can change q and b satisfying the conditions above. However, if you want a fast constant-time implementation, then in any case you need to handle independently all the reminders of q modulo 27 and mutually exclusive cases concerning the cubic residuosity of b.

Yes, please handle both cases in the same file. Sage implementation is far from being constant time and we are expecting that your implementation of the hash function provides the reference and the learning material for those who are going to implement it for production, for example in Rust. As such, it is best served if the clarity of the code is not sacrificed for performance or constant time concerns. I'd leave it to them to worry about this aspect.

> On the other hand, in the real-world cryptography only the curves BLS12-381, BLS12-377 are applied, as far as I know. Therefore, an implementation of the cases 1), 2) should be sufficient for W3F purposes.

I believe we differ on this point. The huge cost of cofactor clearing for these cases leaves us little motivation to invest in a new map to curve function. On the other hand, we think that there might be other cases where the map to curve step is the bottleneck making this hash function is more interesting. Nonetheless, this does not change the fact that a unified implementation class is much more useful for other implementors, as @cruikshankss pointed out.

> In the implementation I stick to the notation of my article. I don't understand how to refer more precisely and how to comment in more detail.

After the implementation is unified in a single class, if it helps, I could add comments for  parts of the code serving as an example of how the code could be explained better and hopefully you could do the same for the rest of the code.

> As regards testing, I cannot test as professionally as it is said in https://doc.sagemath.org/html/en/developer/doctesting.html#chapter-doctesting, because, frankly, this document is difficult for me. But I can only give naive tests, the way I see it.

Sure, tests in [python native unittest framework](https://docs.python.org/3/library/unittest.html) also work for us. I just thought Sage doc tests are easier and more straightforward. Just a note that you need to rename your files from .sage to .py (and import the Sage libraries you are using) if you are using the native testing framework (you could use `sage --preparse foo.sage` for sage to generate the .py file for you).

dishport:
> I believe we differ on this point. The huge cost of cofactor clearing for these cases leaves us little motivation to invest in a new map to curve function. On the other hand, we think that there might be other cases where the map to curve step is the bottleneck making this hash function is more interesting. Nonetheless, this does not change the fact that a unified implementation class is much more useful for other implementors, as @cruikshankss pointed out.

@drskalman, I don't agree with you that cofactor clearing for the subgroup G1 is of the huge cost, because the cofactor is not very big in contrast to the subgroup G2. Moreover, I discuss in Section 1.1 of [my article](https://eprint.iacr.org/2021/1082.pdf) that cofactor clearing for G1 in fact can be completely omitted.

dishport:
> if it helps, I could add comments for  parts of the code serving as an example of how the code could be explained better and hopefully you could do the same for the rest of the code.

@drskalman, it would be great.

drskalman:
> > if it helps, I could add comments for  parts of the code serving as an example of how the code could be explained better and hopefully you could do the same for the rest of the code.
> 
> @drskalman, it would be great.

Please see:
https://github.com/dishport/Indifferentiable-hashing-to-ordinary-elliptic-curves-of-j-0-with-the-cost-of-one-exponentiation/pull/2/files

drskalman:
> @drskalman, I don't agree with you that cofactor clearing for the subgroup G1 is of the huge cost, because the cofactor is not very big in contrast to the subgroup G2. Moreover, I discuss in Section 1.1 of [my article](https://eprint.iacr.org/2021/1082.pdf) that cofactor clearing for G1 in fact can be completely omitted. Thus the map to curve step is actually the bottleneck making my hash function more interesting.

See Page 18 of:

- Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
  hashing to the bls12-381 elliptic curve. IACR Transactions on
  Cryptographic Hardware and Embedded Systems.

The co-factor clearing cost is 53% (592/1121) of the whole hash to G1 operation.

Independant of this fact, there are lots of redundant copy and pasted code in two files (the phi function is the same for both curves for example) which makes the code costly to maintain and follow among other problems. Sage does not guarantee constant time and so there is little motivation to sacrifice code clarity for that.

PR #454, opened 2022-05-30
Tesseract dApp/Wallet Connection: milestone #1

dileping:

- [ ] In case of acceptance, the payment will be transferred to the BTC/ETH payment address provided in the application.

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/508.


dileping:
@Noc2 Thanks for getting back on this. No worries. We are moving full speed ahead with the next milestone. Already have have a working solution on Android and now are integrating it with the Rust Core. Though, we'd really like to move ahead with marketing and publish the articles.

Noc2 (changes requested):
Sorry again for the late reply. I finally started to look into the delivery, but when I run cargo +nightly run, I get the following message: https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/mobile_dapp_connection_1_Noc2.md#general-notes Could you look into this?

dileping:
@Noc2 thanks for starting the review. Already accepted your change to our README.MD - totally forgot about nightly as it was set by default in my environment.

As for the output - it's correct and works properly, displaying two requests/responses done :) one is a successful response, and another one is an intentional error (when the wallet produces an error) made to demonstrate the ability of the protocol to transfer the errors properly.

So here is some explanation:

The output line starting with `@@@WOW@@@` (yes, I was that happy to see the response finally) is printing the response to the request made by line
`let signed = Arc::clone(&client_service).sign_transaction("testTransaction");`
The piece `_signed!` is added by the wallet side and thus we can see that the transfer went through the whole stack properly.

The output line starting with `!!!!UGH!!!!` is the line that prints an intended error. The wallet-side is implemented the way that `make_error` "transaction" is a special case that produces an error for the test purposes.

If you think we need to improve the output and make it more clear, please, let me know. We just didn't bother with it much, but rather kept it understandable for us so that we could test the protocol stack.

Also, I just noticed that in you review the *Application Document* and *Kusama Identity* are probably from different project. Our *Application Document* is https://github.com/w3f/Grants-Program/blob/master/applications/mobile_dapp_connection.md and we didn't provide any *Kusama Identity*.

Noc2 (approved):
Thanks for pointing out that I was linking to the wrong application and already merging my PR. I’m happy to tell you that the milestone is a pass and I will forward your invoice internally. My only comment regarding the above would be that it’s not super user friendly ;-)

Noc2:
Hi. I was referring to the output above: @@@WOW@@@ and !!!!UGH!!!! are not the most common terms ;-)

dileping:
@Noc2 glad to hear it's just that. Thank you!

PR #558, opened 2022-09-01
Blackprint visual programming for Polkadot.js - M3

StefansArya:

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/774


dsm-w3f:
@StefansArya thank you for the milestone submission and my apologies for the late response. Please see [the evaluation document](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/blackprint_3_dsm-w3f.md) and provide proper answers and fixes. After that, let me know to continue with the evaluation.

StefansArya:
Hi @dsm-w3f I also apologies for my delayed response,

Thank you for your great evaluation!
Technically that's what I need to help me improve the quality.

> Some modules have low coverage like Convert (50%), Keyring (60.31%), NodeToast.js (59.72%), Transaction/DryRun.js (39.13%), please review and see if improvements could be performed.

Sure, I think I could do some improvement about it.

> Running unit tests without docker, the tests sometimes pass sometimes fails. I think something is not cleaned between runs because the first time I ran all test passed, then the second time some failed and when I ran again they passed.

Usually the unit test may fail if it was failed to connect with the Westend network, insufficient account balance, or sending a transaction is longer than usual as it has 40sec timeout for each unit test. The WebSocket connection is reused to avoid spamming the network with the client's reconnection for some unit test, I'm not sure it's the problem though. But it will be awesome if you can screenshot the unit test that was failing.

> Is there some browser/plugin requirement for using blackprint? In Firefox on Linux got the error below, using Opera this error didn’t show up.

There's no browser plugin requirement for the Blackprint itself, but the development is mainly focus for Chromium based browser because the HTML/SVG/Canvas rendering is more faster than Firefox and features like PWA didn't exist on Firefox natively. I also was unaware if `computedStyleMap` [isn't compatible](https://developer.mozilla.org/en-US/docs/Web/API/Element/computedStyleMap#browser_compatibility) for Firefox, and I will fix it with `getComputedStyle` [as the alternative](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle). Thank you for reporting about the error.

> Those variables are injected in some way? I think would be nice to review eslint results and fix the warnings that make sense.

Sure, actually every file will be combined into one file but in a ordered way that can be specified on the [configuration](https://github.com/Blackprint/nodes-polkadot.js/blob/1fececa0d17821fa11165f76cb86a5014129eea2/blackprint.config.js#L31-L35). The variable was injected for the hot reload feature and the directory structure is designed to make it possible to view the source code of a node directly from the Blackprint Editor by using JSDelivr's API to get the source files:

Ua8Rk4S7Y3

I have put the commented imports to make it easier to migrate to ES modules and help other developer find where the variables was imported. Some nodes that don't depend on other files can be easily modified inside the editor. But thank you for the warm advice to review the ESLint warning, actually I already considered to use ES module for the source files but I wasn't comfortable developing without hot reload feature. But I will try reconsider it again and if needed I will also create a plugin for developing with other bundlers/compiler.

Please allow me around 2 days to improve the coverage for the module and fix the bug that you have mentioned, and also let me know if I need to do migration to ES modules for this milestone. It can resolve the ESLint warnings, but I may lost the hot reload features and I'm not sure if I can easily support Deno that was importing modules via URL while also supporting Node.js that using local `node_modules`.

StefansArya:
Hi @dsm-w3f, I have done with the improvement.

The ESLint warnings has been resolved and the total coverage was improved to 95%.
You can see the [coverage result here](https://github.com/Blackprint/nodes-polkadot.js/actions/runs/3087847018/jobs/4993646737#step:8:208) for a quick view.

I already fix the bug on Firefox that you mentioned, but it seems I also find other bug to fix for that browser 😅

dsm-w3f:
@StefansArya the coverage looks good and I saw that the tests are passing into the pipeline. However, the unit tests are still failing sometimes for me. I looked into the unit tests code and I discovered that they are integrated with external components such as those addresses 'wss://pangolin-rpc.darwinia.network' and 'wss://westend-rpc.polkadot.io/'. In this way, the unit tests are more likely to be system tests than unit ones since they are integrated with other components. Ideally, unit tests need a controlled environment to run and always produce the same result. Unit tests commonly use mocks and other technologies to guarantee the isolation and reproducibility of the tests. There are some ways to achieve always the same result even though the tests are integrated with persistent state systems (databases, blockchains, etc). This includes setup the database or blockchain to the specific state that you need for running each test before running the test. We also cannot assume that the tests will run in a specific order or allow that one test case changes states that are being used in another one. Reusing addresses, balances, all this influence the way the tests run in an integrated env. There are some specific instructions or states to setup so I would be able to run the tests locally and they pass?

StefansArya:
Hi @dsm-w3f, thanks for the detailed explanation.
I appreciate your thought and will take it into my consideration.
But sadly I'm still not sure if the error message of the failing test is similar like before or not.

Originally I was using Westend to make it easier for us to verify if the transaction is actually working on that network and we can also verify it with Subscan. I'm afraid if I use mocking for the transaction and connection it would never find problem like this. I also think we need to setup local node instead of just mocking the connection or transaction only.

Meanwhile, by looking at the transaction history on [Subscan](https://westend.subscan.io/account/5GqT7pbxcp6FK3EAy9a3nXf9U6YP5ifgCadHhSnqs9vMAWR1) it seems the first transaction was sent to the network from your side, but I didn't see the batch transaction on the history. So it may be something wrong because the other transaction didn't being processed and only one that was processed. After I tried again last night, it seems I finally can reproduce the error with message:

> RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1014: Priority is too low: (12184 vs 5635): The transaction has too low priority to replace another transaction already in the pool.

When testing I never met this error before and think if Polkadot.js library already handle the nonce, and the transaction will never being sent in the same block. Well, without using Westend it may be more difficult for me if you didn't send any related error logs so I can only predict what was wrong or I need to look on the transaction history from Subscan. But by finding this error message I think it can be fixed by manually specify the nonce or waiting for the next block.

---

> There are some specific instructions or states to setup so I would be able to run the tests locally and they pass?

Because you can test it I think you already done it correctly, but only my mistake because I haven't specify the nonce manually and the error happens randomly. In case you was unsure with the setup you can also see the [Dockerfile](https://github.com/Blackprint/nodes-polkadot.js/blob/main/.github/docker/unit-test/Dockerfile). The required states for the current test is having [account balance](https://westend.subscan.io/account/5GqT7pbxcp6FK3EAy9a3nXf9U6YP5ifgCadHhSnqs9vMAWR1) for the test account.

dsm-w3f:
@StefansArya Thank you for the response. I tried again today and all tests passed. The warnings of eslint also were fixed. The milestone delivery is approved, you can check the update in [the evaluation document](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/blackprint_3_dsm-w3f.md). Great job! If you did not send an invoice regarding this milestone delivery feel free to do so.
image

StefansArya:
@dsm-w3f Thank you very much for your evaluation!
It did help me improve with the quality of the module, and at the end I had to remove the DryRun node because not every network supported it (I have tried multiple different networks but then the method got removed after some days and the tests were failing).

About the invoice, I think I had submitted it on the same day when I opened this PR.
But feel free to ask if I need to resubmit again.

PR #592, opened 2022-10-09
Fair Squares M3 delivery

ilhanu:

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/971


semuelle:
Thank you for the delivery, @ilhanu. We will look into it as soon as possible. We currently have a bit of a backlog, so please bear with us. 

As far as I can tell at first glance, there is no need to update the application for one feature and contributor.

alxs:
Hi @ilhanu, thank you also from my side and sorry for the late reply.

Overall, the delivery looks good! It's clear you've made a point of providing great inline documentation and test coverage. Note that the deliveries need to be the same as in the [application](https://github.com/w3f/Grants-Program/blob/master/applications/fair_squares.md#milestone-3---bidding-mechanism) - could you please list the "timer" (which I see is just part of the logic in the bidding pallet) and "fractioning" (which I assume is the share distributor pallet) modules in the delivery file and link to where they're implemented exactly?

Besides, `cargo clippy` again returns a number of warnings and errors, could you address these?

ilhanu:
Thank @alxs for your feedback, we been looking forward to this. 

The Timer was supposed to be a separate module, until we found out about the traits hooks. So it's now defined in the runtime and bidding pallet. After Every Epoch it scans for new status and reserves a part of the housing-fund or bids on an asset. The important code blocks are to be found [here](https://github.com/Fair-Squares/fair-squares/blob/2eb97eb978a4b9350684cb42234ff9bdd6b28473/pallets/bidding/src/lib.rs#L128-L134) and [here](https://github.com/Fair-Squares/fair-squares/blob/9fd6f162e4f4e1e849817e7bdaefd1ee9bec89c7/runtime/src/lib.rs#L1051).  We updated this in the delivery file. 

Concerning the Module:fractioning you are right that it's replaced by the share-distributor-pallet. This functionality became a pallet on it's own as the share-distributor, doesn't just fraction and calculate shares on top of it it also distributes the shares with the asset-pallet to the shareholders, by means of a virtual account. As the complexity grew we decided that this should have its own pallet instead of cluttering the current pallets. The agreed functionality from the fractioning-module can be found [here](https://github.com/Fair-Squares/fair-squares/blob/main/pallets/share_distributor/src/functions.rs#L66-L115) and [here](https://github.com/Fair-Squares/fair-squares/blob/main/pallets/share_distributor/src/functions.rs#L181-L214).  We updated this in the delivery file. 

We ran `cargo clippy` and updated our code based on the warnings and recommendations. We have suppressed some warnings, cause it complains about unused dependencies or values as some parts that clippy thinks are being unused are actually used by tests and are used dependencies.

alxs:
Many thanks for the quick answer and changes. I had a chance to follow the testing guide today and was thrilled by the amount of effort you put into explaining the functionality you developed and making it easy for us to follow along - thank you. The coverage and amount of tests is also superb, as well as the rustdocs and other inline documentation.

Regarding the changes to the deliverables, I'm happy to accept the delivery as is since these are just architectural changes and the scope indeed has not changed. Nevertheless, you may want to [amend](https://github.com/w3f/Grants-Program#changes-to-a-grant-after-approval) the application at some point so it reflects your final architecture, or if you want to deviate from the original milestone plan (in which case you'll need to).

The only problem I ran into was having to go through the creating a new asset + voting process twice since the first time it somehow ended up in the `RejectedForEditing` state. It might have timed out, I'm not sure, maybe you want to look into it. Also, `cargo clippy` still generated a number of warnings (though no more errors) on my machine, I assume you mean you suppressed those warnings locally.

In any case, I'm happy to let you know that the milestone has been accepted! You can find my evaluation notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/fair-squares_3_alxs.md). I'll notify the finance team to pay out the invoice. As usual, please allow for up to 2 weeks of processing time.

ilhanu:
Thank you too for the credit and the valuable feedback, we will amend the the application before we deliver milestone 4. 

Regarding the state of `RejectedForEditing` it most likely got timed out as the votingPeriod is currently set to 3 mins for the council and democracy. This is not a final parameter, but a current development parameter for us to test in Polkadot.Js. What we did not include in the guide is that you can propose the same proposal nr. with an extrinsic if it did not get addressed by the council, hence the state `RejectedForEditing`

Regarding the `cargo clippy `we do need to look into this more during our development and maybe even use it as we add PR's instead of doing it as final batch per milestone. We indeed suppressed the warnings what we were clueless with as I mentioned some of the dependencies and variables were being used. We need to take a deeper dive into this.

PR #596, opened 2022-10-12
Add instructions to keep RFP in sync

takahser:


Noc2 (approved):
Good idea to add it here

PR #644, opened 2022-12-01
Calamar milestone 1 delivery

uiii:

- [X] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [X] This pull request is being made by the same account as the accepted application.
- [X] In case of acceptance, the payment will be transferred to the BTC/ETH payment address provided in the application.
- [X] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/Grants-Program/blob/master/docs/milestone-deliverables-guidelines.md).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/1163

---------------------------

~~**IMPORTANT** we put a wrong BTC address in the application, the correct one is `32NoFbB4x8bZ7YNvjra1DUYcje2B2XQwP3` (is also mentioned in the invoice). Should we update it in the application as well?~~ **Fixed**

### DIFF
--- a/deliveries/calamar-milestone_1.md
+++ b/deliveries/calamar-milestone_1.md
@@ -1,0 +1,35 @@
+# Milestone Delivery :mailbox:
+
+**The [invoice form :pencil:](https://docs.google.com/forms/d/e/1FAIpQLSfmNYaoCgrxyhzgoKQ0ynQvnNRoTmgApz9NrMp-hd8mhIiO0A/viewform) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/milestone-deliverables-guidelines.md).**
+
+* **Application Document:** https://github.com/w3f/Grants-Program/blob/master/applications/Calamar.md
+* **Milestone Number:** 1
+
+**Deliverables**
+
+| Number | Deliverable | Link | Notes |
+| -----: | ----------- | ---- | ------ |
+| 0a. | License | [LICENSE](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/LICENSE.txt) | |
+| 0b. | Documentation | [README](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/README.md), [Calamar docs](https://docs.calamar.app/) | |
+| 0c. | Testing Guide | [README#Test](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/README.md#test), [E2E tests](https://github.com/topmonks/calamar/tree/c6b96c2c940149a13f67272031af36aecf349e93/test/e2e), [GitHub workflow](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/.github/workflows/test-and-deploy.yml#L27), [Test reports](https://github.com/topmonks/calamar/deployments?environment=test-report#activity-log), [Screenshots](https://app.argos-ci.com/topmonks/calamar) | |
+| 0d. | Docker | [Dockerfile](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/Dockerfile), [README#Docker](https://github.com/topmonks/calamar/blob/c6b96c2c940149a13f67272031af36aecf349e93/README.md#docker) | |
+| 0e. | Article | [Article](https://medium.com/topmonks/calamar-block-explorer-milestone-1-finished-93f683ddc486) | |
+| 1. | Fix usage blockers | [PR#16](https://github.com/topmonks/calamar/pull/16), [PR#41](https://github.com/topmonks/calamar/pull/41) | |
+| 2. | Add related items listings | [PR#66](https://github.com/topmonks/calamar/pull/66), [Example](https://fad15198.calamar.pages.dev/acala/block/0002408225-e923b) | |
+| 3. | Add call detail page | [PR#47](https://github.com/topmonks/calamar/pull/47), [Example](https://fad15198.calamar.pages.dev/kusama/extrinsic/0015560843-000002-1563b) | |
+| 4. | Add event detail page | [PR#47](https://github.com/topmonks/calamar/pull/47), [Example](https://fad15198.calamar.pages.dev/acala/event/0002408225-000008-e923b) | |
+| 5. | Responsiveness | [PR#75](https://github.com/topmonks/calamar/pull/75) | |
+| 6. | Extrinsics/event args displaying improvements | [PR#84](https://github.com/topmonks/calamar/pull/84), [PR#91](https://github.com/topmonks/calamar/pull/91), [Example](https://fad15198.calamar.pages.dev/kusama/extrinsic/0015560839-000001-ae6b8) | |
+| 7. | Items count | [PR#65](https://github.com/topmonks/calamar/pull/65/files) | |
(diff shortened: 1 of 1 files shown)


alxs:
Hi @uiii, thank you for the delivery. I look forward to diving into it as soon as possible. Yes, in that case please [update](https://github.com/w3f/Grants-Program#changes-to-a-grant-after-approval) the BTC address in the application.

uiii:
@alxs great, thanks, like to hear you feedback.

> Out of all things, I seem to have trouble with the line endings when trying to build the Docker image: [docker-build.txt](https://github.com/w3f/Grant-Milestone-Delivery/files/10141347/docker-build.txt). Could you look into it?

I will fix it.

> I also noticed some searches seem to [time out](https://7ffa6aa5.calamar.pages.dev/polkadot/search?query=XcmPallet.reserve_transfer_assets).

Yes, we are aware of some cases where this doesn't work. This require the change of the backend system which is in control of Subsquid guy. We are discussing it with them, so we are definitely going to fix that, but it require more time to test new solutions and adapt our app to them.

> the remark itself could take up more space

Idealy yes, but it is not easy to done, because I need to have the parsed view consisted for all the cases. It contains nested tables and I set the min-width of keys and values. It is compromise between the condensed view and the scrollbar. But I can think more about it. I thought about recalculating the sizes in runtime using javascript so it get the ideal layout.

> it's not clear what System and remark stand for or why the latter is listed twice

It might be confusing, but it makes sense if you check the "raw" data. The `System` and first `remark` represnt the name of the call `System.remark` and the second `remark` is a key from the arguments to that call. I dont se a better way to display it (actually I do, but it requires to understand the semantics of the data, which is very complex thing).

> I would also add page numbers to list views.

Yes good point. That kind of pagination remained here from times where we did not know the total counst of items.

> By the way: If you want us to retweet an announcement linking to the article, feel free to send an email to [grantsPR@web3.foundation](mailto:grantsPR@web3.foundation).

Ah great. Can we (or others if retweeting) mention W3F grant in our tweets?

uiii:
the fix for docker build is in master

alxs:
Thanks a lot for your answers and for the fix. I confirm Docker is working now.

For deliverable 2, it seems that while both extrinsics and calls are [displayed](https://7ffa6aa5.calamar.pages.dev/acala/block/0002408225-e923b) in a block's detail page, transfers are not. Could you add transfers as described in the application? What is the difference between calls and extrinsics, and is it really helpful to display both? The extrinsic detail page also [seems to](https://7ffa6aa5.calamar.pages.dev/acala/extrinsic/0002408225-000002-e923b) always link to a call with the same ID. This distinction doesn't exist for example [on Subscan](https://acala.subscan.io/extrinsic/2447389-2), where "Call" only represents the function that was executed. 

By the way, I noticed accounts are currently displayed in hexadecimal instead of using the correct [SS58 format](https://wiki.polkadot.network/docs/learn-account-advanced#address-format). Do you want to implement this now, or were you planning to address it in a later milestone?

> It might be confusing, but it makes sense if you check the "raw" data. 

It might make sense if one checks the "raw" data, but it's confusing ;) In UX terms, I think it makes more sense to think about it this way. But I understand it may be a complex task and you have other things to focus on. Again, just feedback.

> Ah great. Can we (or others if retweeting) mention W3F grant in our tweets?

Yes, but we ask teams to wait until the first milestone has been accepted. Please have a look at our [Announcement Guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/announcement-guidelines.md).

uiii:
> For deliverable 2, it seems that while both extrinsics and calls are [displayed](https://7ffa6aa5.calamar.pages.dev/acala/block/0002408225-e923b) in a block's detail page, transfers are not.

It was meant block's extrinsics not transfers, I don't know why we wrote "tranfers" here 😕 We are not currently able to detect which extrinsic is transfer.

> What is the difference between calls and extrinsics, and is it really helpful to display both? The extrinsic detail page also [seems to](https://7ffa6aa5.calamar.pages.dev/acala/extrinsic/0002408225-000002-e923b) always link to a call with the same ID. This distinction doesn't exist for example [on Subscan](https://acala.subscan.io/extrinsic/2447389-2), where "Call" only represents the function that was executed.

It was suggested by Subsquid as they have both entites in their archives. Call entity represents extrinsics action, there are cases where it is useful to have the distiction between the two, e.g. batch calls. See [this extrinsic](https://calamar.app/kusama/extrinsic/0015640851-000002-1b90b), it has `Utility.batch` call which has two child calls. 

> By the way, I noticed accounts are currently displayed in hexadecimal instead of using the correct [SS58 format](https://wiki.polkadot.network/docs/learn-account-advanced#address-format). Do you want to implement this now, or were you planning to address it in a later milestone?

Can you give me a link where it uses hex format, please? Because a I know we use SS58 encoding where possible

> It might make sense if one checks the "raw" data, but it's confusing ;) In UX terms, I think it makes more sense to think about it this way. But I understand it may be a complex task and you have other things to focus on. Again, just feedback.

Yeah I understand it may be confusing, current implementation is the best we've come up with. Maybe we will find a better way later.

> Yes, but we ask teams to wait until the first milestone has been accepted.

Yes, I will wait with it.

uiii:
And the IDs (of extrinsic, calls, ...) are internal values from Subsquid, but they keep a specific pattern which can be mapped to be compatible with subscan, but it need a little bit more research.

uiii:
> could you please submit an [amendment](https://github.com/w3f/Grants-Program#changes-to-a-grant-after-approval) and move it to a different milestone (from the [Account Detail](https://github.com/w3f/Grants-Program/blob/master/applications/Calamar.md#account-detail) section and the deliverables)?

The block's "transfers" instead of "extrinisics" in the first milestone deliverable is more like a typo. The [Account Detail](https://github.com/w3f/Grants-Program/blob/master/applications/Calamar.md#account-detail) describes the final application state and is part of the second milestone, so the transfers here are correct. But I can update the application and fix the typo.

> Ah, of course! That makes sense. Nevertheless this seems useful for that specific case, but in the case of all other extrinsics or e.g. listing all extrinsics and calls in the block details page seems superfluous.

Yes it seems superfluous a little bit, I will think about it. Maybe we can have the call list in extrinsic's detail page only.

> It might only apply to specific pages then. I noticed it [here](https://calamar.app/polkadot/extrinsic/0013202686-000002-b65d6) (both for the account and the destination).

Ah it was actually a bug, all the addresses in polkadot relay chain entities wasn't encoded, it is fixed in the master now and released into production.

alxs:
Hey @uiii, sorry for the late reply. I have a bit of a backlog.

> The [Account Detail](https://github.com/w3f/Grants-Program/blob/master/applications/Calamar.md#account-detail) describes the final application state and is part of the second milestone, so the transfers here are correct. But I can update the application and fix the typo.

I see, so you will actually implement it in the next milestone for the Account Details view. Even so, I think it'd be good if you amended it for this milestone, since the deliverable in milestone 2 doesn't cover the same functionality. You can also change deliverable 8 so it matches the implementation, and make changes to the upcoming milestones if you need to. It shouldn't take long to be approved.

uiii:
All right @alxs, I made an amendment.

alxs:
Thank you. With this, I'm happy to let you know that the milestone has been accepted! You can find my evaluation notes [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/calamar_1_alxs.md). 

I'll notify the operations team to pay out your invoice. Please allow for up to 14 days for processing.

uiii:
Great @alxs, I'm happy to hear that. Thanks

RouvenP:
hi @uiii we transferred the payment yesterday

PR #678, opened 2023-01-06
Update README.md

takahser:
Since USDT on Statemint/Statemine is supported, applicants can also provide their DOT/KSM addresses for payment.


Noc2 (approved):
Thanks!

PR #735, opened 2023-02-08
Deliverable: Research Feasibility of a Java Host

Daniel-K-Ivanov:

- [X] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [X] This pull request is being made by the same account as the accepted application.
- [X] In case of acceptance, the payment will be transferred to the BTC/ETH payment address provided in the application.
- [X] The delivery is according to the [Guidelines for Milestone Deliverables](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/milestone-deliverables-guidelines.md).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/1353


semuelle:
Thanks for the delivery, @Daniel-K-Ivanov. I will look into it as soon as possible.

bhargavbh (commented):
**Overall Evaluation:**
The feasibility research has been performed in sufficient detail with clear findings (positive/negative) for each of the subcomponents. The report is well structured and is a good starting point for teams to build a Java node implementation, and nicely complements the spec. 
As mentioned in the grant application, a PoC for testing the performance and compatibility of a Wasm Runtime within Java has been delivered. The performance on small-scale benchmarks are promising. Milestones outlined in the application have been satisfactorily delivered. 

I have looked into the light client design in details and here are a few comments.
**Evaluation of [Java Light Client Design](https://github.com/LimeChain/java-host-research/blob/main/research/java-host-light-client-high-level-design.md)**:

- Network: Are there any optimisations/ tunings planned for peer discovery of light-clients in comparison to Full nodes. The requirements and settings are different, and I would assume implementing the same peer discovery for both cases would lose out on intent-specific optimisations. By design, light-clients peer with fewer nodes in comparison to full nodes. In optimal cases, syncing with a single honest node is sufficient. However, it usually has to sync with at least a few when initially joining the network (due to weak subjectivity). 

- Database: could you please add more details on how you plan to use the persistence layer to store checkpoints and reliable nodes to sync with? This might be a potential attack surface. Our current light-client architecture does not provide a channel to report equivocation back to the network, which means the full nodes can attempt to mislead the light-client without being reported. 

- Host Configuration: The Diagram in the document has a sub-component called `Host Configuration`. What are the parameters of this configuration? Why does it sit between the `network` and `core` layers? I was unable to find any details in the text.

vikinatora:
Hi @bhargavbh, thanks for the feedback and the comments. I'll try to address them here:

1. The primary focus of the research was the feasibility part of the host as well as providing a high-level design of the different host types. Algorithm optimizations weren’t research topics we devoted time to. That’s not to say we don’t intend to deliver on a high-performing client, we intend to modularize the light client in such a way that it’s going to be easy to change implementation details if required for optimization or other reasons.

2. I assume the attack you’re talking about is the long-range attack mentioned in the specification. Our understanding is that this attack is high-risk, high-investment and as such we believe the chance of happening is low and the impact would be low as well since the light client is a passive participant of the network. As you said the punishment for this is reporting the equivocations and light clients don’t have a way to do that. 

   We couldn’t find suggestions neither in the spec nor in the existing node implementations for approaches for guaranteed protection against such an attack or selecting a reliable full node when warp syncing. That’s why we think it’s best to safe guard the code in terms of failure-resistant logic if this attack was to happen. 

    An approach we thought of is to sync with a node until it gives us a reason not to by detecting equivocations using the [justification message definition](https://spec.polkadot.network/#defn-grandpa-justification). We can then move to a new node and  persist the old one to a “black list” of nodes that we’d avoid syncing with anymore and the new one’s address to the “reliable list”. This approach might seem naive but it's aligned with the goal to sync with as few nodes as possible while keeping the complexity low and allows to incrementally improve the functionality.

3. The Host Configuration is small component that is to store host specific information such as the client parameters that it can launch with, chain specifications and other information that may be host specific. The positioning of the component between the network and core layers, doesn’t have to do so much with it’s functionality, rather that it’s the place where we decided is most suitable for that module when talking about diagram symmetry, although you can make the case the configurations will have some implications on the neighbouring layers. It didn’t deserve a full row like the network and core layers because it should be much smaller in terms of logic, but also we had to include it as it’s going to be a light client module after all.

We're open to suggestion for any of these points. Even though we've conducted a thorough research into the inner workings of the host, we're still far from the knowledge that people at Parity or the host's core developers have. Our team hopes that one day we'll be at the tip of the knowledge in the Polkadot community as well.

vikinatora:
Hey @bhargavbh, @semuelle. Have you managed to take a look at our reply to your comments? Do we need to update the reasearch outcome documents or the discussion will be kept here?

bhargavbh:
hi @vikinatora thank your for the replies. Of course, comments-1&2 were exploratory, however, i would suggest integrating your clarification on Host Config in the research doc. My colleague @FlorianFranzen is evaluating the PoC WASM environment deliverable, and will reply here soon. 

Cheers

vikinatora:
Thanks for the info @bhargavbh, I updated the Light Client design document to have information about the Host Configuration module. Looking forward to hearing from you.

Noc2 (approved):
Sorry for the delay here. I'm finally happy to tell you that the milestone is approved and I will forward your invoice internally. [See our notes here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/LimeChain_Java-Host-Research_M1_bhargavbh.md)

fededubbi:
Hi @vikinatora,

Can you please let us know in which currency you want to be paid in?

Many thanks,
Federica

vikinatora:
Hi @fededubbi,

Address is eth:0x1F7683228Ee9Bc65335374eA5c92B81C74131404 (USDC/USDT/DAI).

Thank you,
Viktor

vikinatora:
Hey @fededubbi,

DAI is preferred.

Thanks,
Viktor

RouvenP:
hi @vikinatora we transferred the payment today. we had to do the transfer in USDC due to liquidity constraints. thanks!

PR #750, opened 2023-02-14
Update check_author.yml

0xCaso:
Workflow triggered also if PR is updated, otherwise if one opens the PR and instantly updates it, the check won't be done a second time.

### DIFF
--- a/.github/workflows/check_author.yml
+++ b/.github/workflows/check_author.yml
@@ -1,7 +1,7 @@
 name: Check Author
 on:
   pull_request_target:
-    types: [opened]
+    types: [opened, edited]
 
 jobs:
   get-delivery-files:
--- a/.github/workflows/check_author.yml
+++ b/.github/workflows/check_author.yml
@@ -1,7 +1,7 @@
 name: Check Author
 on:
   pull_request_target:
-    types: [opened, edited]
+    types: [opened, synchronize]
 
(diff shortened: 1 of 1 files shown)


semuelle (commented):
Can you show an example of where this would be needed? Unless I'm misreading this action, editing the PR wouldn't change its author, as that's the person clicking the `Submit pull request` button. Or can more than one person edit a PR?

Btw, if by 'updates' you are referring to making commits to the PR, that wouldn't be addressed with `edited`, that would be `synchronize`. I _thnk_ `edited` only triggers when the title or body of the PR is edited, not the underlying files.

0xCaso:
Thanks yes, i was meaning `synchronize`. Hope that this makes sense

semuelle:
But a commit to a PR doesn't change the PR author. If I added a commit to this PR right now, it would list the two of us as commit authors but still report you as the PR author, which is what the action is using. So nothing has changed from the action's point of view. Or not?

0xCaso:
The thing is this:
1. X opens the application PR and it gets merged
2. Y opens the delivery PR
3. the workflow says there is an issue (Y!=X)
4. Y pushes a change to the PR
5. the workflow won't re-run, and there is no check that Y == X so potentially all the checks will pass (with `synchronize` the check would run also in this case)

PR #754, opened 2023-02-17
Update badge_or_last_milestone.yml

semuelle:
See https://github.com/w3f/Grant-Milestone-Delivery/actions/runs/4206981381.

Not tested.


0xCaso (approved):
Thanks @semuelle for pinging me

We have to check that the delivery is not a maintenance delivery, because in that case we don't want to send the message (also because we don't know, in theory, how many milestones the grant will have)

PR #887, opened 2023-06-20
Dotflow Milestone 1 delivery

Szegoo:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [x] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [x] This pull request is being made by the same account as the accepted application.
- [x] I have disclosed any and all sources of reused code in the submitted repositories and have done my due diligence to meet its license requirements.
- [x] In case of acceptance, the payment will be transferred to the BTC/ETH/fiat account provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://grants.web3.foundation/docs/Support%20Docs/milestone-deliverables-guidelines).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/1657

PR #888, opened 2023-06-20
External evaluation of Dotflow Milestone 1

dastansam:
External evaluation for the Dotflow Milestone 1

- Delivery [PR](https://github.com/w3f/Grant-Milestone-Delivery/pull/887)
- Application [PR] (https://github.com/w3f/Grants-Program/pull/1657)


Szegoo:
@dastansam Thanks a lot for your evaluation! I have added instructions regarding [contract deployment](https://github.com/TheDotflow/dotflow-ink#deployment) and [frontend environment variable](https://github.com/TheDotflow/dotflow-ui#set-up-development-environment).

I have also [resolved](https://github.com/TheDotflow/dotflow-ink/pull/37) all the clippy errors, thanks for pointing that out.

Regarding `e2e-tests`, yes we were planning to add them once we have all the pieces of the logic.

The reason why we use `assert` or `unwrap` in some parts of the code is because we have ensured earlier that the unwrap will be successful(like [here](https://github.com/TheDotflow/dotflow-ink/blob/3ec6be1712b8074352277ee2f23e0d3c6faa0196/contracts/identity/lib.rs#L254-L256)).
Also, using `assert` and `unwrap` in the constructor is in my opinion ok since it is not a problem if the code panics during the contract deployment, but if that is a problem we can update the code.

dastansam:
hey @Szegoo,

thanks for the swift feedback.

> The reason why we use assert or unwrap in some parts of the code is because we have ensured earlier that the unwrap will be successful(like [here](https://github.com/TheDotflow/dotflow-ink/blob/3ec6be1712b8074352277ee2f23e0d3c6faa0196/contracts/identity/lib.rs#L254-L256)).

yeah, I went through it again and you are right, `unwrap` calls are safe. But in this case, I would suggest a cleaner syntax, something like this:

```rust
// instead of this
ensure!(self.number_to_identity.get(receiver).is_some(), Error::IdentityDoesntExist);

let receiver_identity = self.number_to_identity.get(receiver).unwrap();

// do this			
let receiver_identity = self.number_to_identity.map_err(|_|, Error::IdentityDoesntExist)?;
```

> Also, using assert and unwrap in the constructor is in my opinion ok since it is not a problem if the code panics during the contract deployment, but if that is a problem we can update the code.

I think in general, it's not that critical to use `unwrap` and `assert` in the contract env, compared to runtime or client env, where it is unacceptable. But it's still better to avoid it, IMO.

But overall, again, I don't see any critical issues that needs to be addressed.

Szegoo:
@dastansam Yes, you are right. It is a good idea to refactor the code the way you described. Thanks for your suggestion. https://github.com/TheDotflow/dotflow-ink/pull/38

dastansam:
It's good to go for me, let's wait for @keeganquigley

keeganquigley:
Thanks for the evaluation @dastansam excellent work! @Szegoo I agree, everything looks great and I am able to reproduce the results. A couple of comments:

- I agree that it's nice to have the process of deploying the contract on a local testnet included for those who may not have done it before. Thanks for adding it and improving the docs overall. 
- I agree, no worries about the e2e tests until the next milestone.
- @dastansam could you change the status to accepted? Once that is done I will merge it along with the delivery.

Thanks!

dastansam:
@keeganquigley changed the status of the evaluation to `Accepted`

keeganquigley:
Thanks a bunch @dastansam I also made a minor title fix to adhere to our naming convention. I will forward your KSM address to our operations team to process your payment.

RouvenP:
hi @dastansam we just transferred the KSM

PR #920, opened 2023-07-12
milestone 2 supersig_fellowship

decentration:
# Milestone Delivery Checklist

- [x] The [milestone-delivery-template.md](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/deliveries/milestone-delivery-template.md) has been copied and updated.
- [x] The [invoice form :pencil:](https://forms.gle/LSRr7PCjBpEbKGh89) has been filled out for this milestone.
- [x] This pull request is being made by the same account as the accepted application.
- [x] I have disclosed any and all sources of reused code in the submitted repositories and have done my due diligence to meet its license requirements.
- [x] In case of acceptance, the payment will be transferred to the BTC/ETH/fiat account provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://grants.web3.foundation/docs/Support%20Docs/milestone-deliverables-guidelines).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/1630 < please fill this in with the PR number of your application.

PR #923, opened 2023-07-13
External evaluation of Supersig UI milestone 2

dastansam:
Overall, everything except the tests on a local chain is working as expected. 

Milestone delivery PR: #920 
Application: https://github.com/w3f/Grants-Program/blob/master/applications/supersig_fellowship.md


decentration:
Hey @dastansam, thanks for the swift eval. Oh I know what the problem is. It’s because a different version of the local node is using an older version of pallet_supersig, which uses `propose_call` instead of `submit_call` that’s why submit call is not found. It’s probably because I shared you a link to substrate_supersig_template which is leading to the incorrect branch… the branch should be `v0.9.37-fix-unbounded`. 

Ah, no actually, I didn’t push the changes to the repo. If you recompile again (perhaps also cargo update), then rerun the chain, then it should be smooth sailing after that!

dastansam:
`cargo update` fixes it! thanks for quick response @decentration!

Successful local tests

dsm-w3f:
@dastansam thank you for the external evaluation. It is accepted. I'll forward your KSM payment. Great Job!

PR #953, opened 2023-08-01
Update README.md

semuelle:
Limit external evaluations to one at a time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Grant applications can be [amended](https://grants.web3.foundation/docs/Process/
 
 ## :ballot_box_with_check: External Evaluations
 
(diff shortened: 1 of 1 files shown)


keeganquigley (approved):

Whisker17:
I would suggest limiting it to one draft PR at a time and once the external evaluation has been set to ready to merge, the next external assessment can be applied for, wdyt?

dsm-w3f (approved):

semuelle:
Thanks for the input, @Whisker17. For now, we'll use this simpler rule and see how it goes. If it proves too much of a burden, we'll implement something to the effect you suggested.

PR #1013, opened 2023-09-26
Delivery of the 2nd milestone (TPScore)

its-a-setup:

- [x] I have disclosed any and all sources of reused code in the submitted repositories and have done my due diligence to meet its license requirements.
- [x] In case of acceptance, an invoice must be submitted and the payment will be transferred to the BTC/ETH/fiat account provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://grants.web3.foundation/docs/Support%20Docs/milestone-deliverables-guidelines).

Link to the application pull request: [1837](https://github.com/w3f/Grants-Program/pull/1837)

### DIFF
--- a/deliveries/tpscore_milestone_2.md
+++ b/deliveries/tpscore_milestone_2.md
@@ -1,0 +1,37 @@
+# Milestone Delivery :mailbox:
+
+**The [invoice form :pencil:](https://docs.google.com/forms/d/e/1FAIpQLSfmNYaoCgrxyhzgoKQ0ynQvnNRoTmgApz9NrMp-hd8mhIiO0A/viewform) has been filled out correctly for this milestone and the delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/milestone-deliverables-guidelines.md).**  
+
+* **Application Document:**  [TPScore Application](https://github.com/w3f/Grants-Program/blob/master/applications/TPScore.md)
+* **Milestone Number:** 2
+
+**Context** 
+
+TPScore simplifies TPS Data Analysis for non-technical users in the Polkadot Ecosystem.
+
+TPScore aims to provide non-technical users in the Polkadot ecosystem with an accessible and user-friendly platform for analyzing TPS (Transactions per Second) data. Our goal is to bridge the gap between technical intricacies and user-friendly visualization, empowering individuals to make informed decisions about blockchain adoption, investment, and development.
+
+The project consists of two main parts:
+
+1. ETL System
+2. User Interface (UI) with Data Visualizations
+
+This milestone contains the User Interface (UI) with Data Visualizations according to the original [TPScore grant application](https://github.com/w3f/Grants-Program/blob/master/applications/TPScore.md).
+
+**Deliverables**
+
+After submission of this milestone all changes to the repo will happen outside of 'main' branch.
+
+| Number | Deliverable | Link | Notes |
+| ------------- | ------------- | ------------- |------------- |
+| **0a.** | License | [Licence](https://github.com/BCS-Labs/tpscore-web-client/blob/main/LICENSE) | GPLv3|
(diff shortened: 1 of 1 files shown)


keeganquigley:
Thanks for the delivery @its-a-setup and sorry for the delay. You can find my evaluation [here](https://github.com/w3f/Grant-Milestone-Delivery/blob/master/evaluations/tpscore_2_keeganquigley.md).I just have two questions about the stats. 

- What exactly do these tx/s stats mean? Many of them say 0.0 tx/s, which doesn't really provide any helpful information. I'm wondering if the data format should be changed.
- How often will these feeds be updated? I see the latest feeds on the published site were last updated on August 9th.

its-a-setup:
Hi @keeganquigley 
Thanks for the evaluation! Happy to see that it works

Below are the answers:
1. tx/s is the number of transactions per second (TPS). We calculate the average TPS based on the last 100 blocks and include only extrinsics of type "Balances". 0.0 tx/s is displayed for some of the parachains due to low on-chain activity happening there. This is also not ideal from our end, so we are thinking about including transactions of different types in the calculation in the next release. 
2. The data is updated every 10mins based on the last 100 blocks. We planned to turn it on after the official announcement together with Web3 Foundation, but I just turned it on, so you could check that it works :)

keeganquigley:
Thanks @its-a-setup for the explanation and for updating the feeds. I'm still curious about how benchmarking was performed for these results. I know TPS metrics are subjective, but as this [wiki article](https://wiki.polkadot.network/docs/faq#what-is-polkadots-transactions-per-second-tps) notes, Polkadot and substrate-based chains should be able to achieve somewhere over 1000 tx/s for simple balance transfers. I know that some parachains might not have much activity, but the Polkadot relay chain itself has a [lot of activity](https://polkadot.js.org/apps/#/explorer). So I would think at least this metric would be higher. 

Are these benchmarks accounting for the different block times of each chain? What about latency? I agree it might help to include other extrinsics than just transfers, especially for chains that don't have a lot of transfers happening. Any further explanation you can give regarding the results and how you got them would be great. Thanks!

its-a-setup:
Sure, let's dive deeper into that! 
This [wiki article](https://wiki.polkadot.network/docs/faq#what-is-polkadots-transactions-per-second-tps) relies on technical max throughput, although the situation on-chain may be different, due to lower utilization and a smaller number of users. This is not only the case for Polkadot though, most of the other L1s have the same thing. E.g. Solana theoretically can handle 710,000 transactions per second, but in reality, it handles just ~230 transactions per second. I would say the fact that on-chain activity for Polkadot Relay Chain and parachains is much smaller than theoretical throughput is rather the norm, which is also observed on other chains.

As for the Relay Chain specifically, we can do a quick check and take a look at the last 20 blocks  (e.g. [17598627](https://polkadot.subscan.io/block/17598627) - [17598646](https://polkadot.subscan.io/block/17598646)) and we will see that at max blocks have 2 transactions of type "Balances" (e.g. in block [17598637](https://polkadot.subscan.io/block/17598637)). In total these 20 blocks have 6 transactions of type "Balances", considering that the time between blocks is 114 seconds, it will mean that TPS is 0.05, which is perfectly in line with calculations based on 100 blocks.

>Are these benchmarks accounting for the different block times of each chain?

Irrespective of block time, we still take 100 blocks for any chain

>What about latency? 

We don't consider it at the moment

keeganquigley:
Thanks for the clarifications @its-a-setup much appreciated. I still recommend perhaps displaying more decimal places or something like that, but I'm happy to go ahead pass the milestone. Nice work! Please allow for up to 14 days for processing.

its-a-setup:
@keeganquigley thank you! 
This is amazing news and a huge milestone for us! 

As for the 0 TPS values for some of the chains, it will be one of our core focuses moving forward. It's also super important to make TPS calculation consistent between parachains and blockchain networks outside of the Polkadot ecosystem so that dApp builders could get a better insight on which parachain they should build. We will try to make an improvement on that in the 2nd version of our app.

PR #1090, opened 2023-12-23
calamar M3 evaluation by PieWol

PieWol:
External evaluation for the 3. milestone of calamar.


semuelle (approved):
Thanks for the evaluation, @PieWol. I can confirm that everything works as expected. As for the metadata, I agree that it would be nice to have, but as Richard pointed out, it's out of scope for this milestone. 

Your evaluation is hereby accepted. Someone will be in touch regarding your KSM soon.

PR #1114, opened 2024-01-22
Move QSTN milestone 1 delivery to `deliveries` directory

davidsemakula:
Moves `QSTN-milestone_1.md` from the root directory to the `deliveries` directory


semuelle (approved):
Thanks!

PR #1117, opened 2024-01-24
DAuth 2 eval

PieWol:


yyd106:
Hi, @PieWol 
Few updates here:
1. SDK license has been added;
2. The inline documentation is in our code, as you can find them when you review our delivered repos;
3. The SDK wallet has been integrated into Hexlink: https://www.hexlink.io/. They use DAuth as the email authentication;

PieWol:
Hey @yyd106 ,
thanks for adding the SDK license. I could also see that hexlink actually integrated your code even though it seems like an abandoned project since last year. 

I could see some inline documentation but I would appreciate it if you could extend it slightly. E.g. the [encrypt function](https://github.com/DAuth-Network/dauth/blob/c759d61b51aff3ff5d11a4a41d670236bd739dbb/packages/core/src/utils/crypto.ts#L5) is without any comments but seems to be a key component of your code. As soon as that's done I see all issues resolved. As this is the final milestone it would be a good opportunity to put in a final effort to simplify further development by others. 

Thanks for your effort!

yyd106:
Hi @PieWol , the inline comments have been added accordingly.

In addition, the Hexlink team is working on some new projects and has also promised to use DAuth/Openid3.

PieWol:
Hey @yyd106 ,
I can't seem to find the added inline docs, would you mind linking the updated version to me? Looks like the last commit to main was the addition of the license. Also thanks for updating me on the hexlink team. I'm glad to hear that they are still integrating with you. 🎉

yyd106:
Hi @PieWol , The update is made on the grants branch. You can find the changes here: https://github.com/DAuth-Network/instant-proof/compare/main...polka-grant

PieWol:
I see, thanks for linking it and putting in the final effort. I updated my evaluation. @yyd106

semuelle (approved):

PR #1166, opened 2024-04-09
Update badge_or_last_milestone.yml

keeganquigley:
Fixed version typo to resolve this issue: https://github.com/w3f/Grant-Milestone-Delivery/actions/runs/8609197620/job/23593781417

Also proposing that we remove `check-needs-badge` as a dependency so that `request-invoice` job can be independent, since we want it to fire regardless of badge. (It still checks to make sure delivery files are present).


semuelle (approved):
Thanks!

PR #1201, opened 2024-07-30
Cyborg Network M1 eval

ditavia-br:


beekay2706:
Hi @ditavia-br,

Thank you for the detailed evaluation and constructive feedback.

Regarding the Docker setup, we were unable to create a single Docker image to run the entire system because the setup requires connecting multiple VMs to test the full system flow. Deploying all components (Substrate chain, K3s master node, and K3s worker) on the same VM is not feasible.

I apologize for the inconvenience. As the next step, we will focus on creating a UI for adding K3s clusters through the frontend under the "Provide Compute" section.

You can use any Polkadot wallet to test this flow. By default, the chain only populates balance for the Alice keypair, which is why you might see the "account balance too low" error during testing. We’ve updated the instructions to include steps for adding the Alice keypair, which comes with a preloaded balance by default.

Please ensure you follow the networking instructions in the testing guide to accurately recreate the deployment of K3s workers.

I hope the updated instructions are helpful. Given the complexity of the system, we would be happy to connect via Element or Zoom if you need further clarification.

keeganquigley (approved):
Thanks @ditavia-br we appreciate it as always!

PR #1278, opened 2025-08-28
Dotnix v2

Ra33it0:

- [x] I have disclosed any and all sources of reused code in the submitted repositories and have done my due diligence to meet its license requirements.
- [x] In case of acceptance, invoices must be submitted and payments will be transferred to the Polkadot AssetHub and/or fiat account provided in the application.
- [x] The delivery is according to the [Guidelines for Milestone Deliverables](https://grants.web3.foundation/docs/Support%20Docs/milestone-deliverables-guidelines).

Link to the application pull request: https://github.com/w3f/Grants-Program/pull/2492

### DIFF
--- a/deliveries/dotnix_followup_milestone-1.md
+++ b/deliveries/dotnix_followup_milestone-1.md
@@ -21,6 +21,6 @@ defined policies, providing an additional layer of security beyond traditional d
 | 0a. | License | https://github.com/sporyon/dotnix-core/blob/main/LICENSE |
 | 0b. | Documentation | https://github.com/sporyon/dotnix-core/blob/main/README.md |
 | 0c. | Testing and Testing Guide | https://github.com/sporyon/dotnix-core/tree/main/checks | 
-| 0d. | Docker | https://github.com/sporyon/dotnix-core/blob/main/docker.nix |
+| 0d. | Docker | https://github.com/sporyon/dotnix-core/blob/main/example.nix |
 | 1. | SELinux | https://github.com/sporyon/dotnix-core/tree/main/pkgs/selinux |
-| 2. | Usability | | 
+| 2. | Usability | https://github.com/sporyon/dotnix-core/commit/ddff867f8a6c4aa80778abd06f5312d5181adce3 | 
--- a/deliveries/dotnix_followup_milestone-1.md
+++ b/deliveries/dotnix_followup_milestone-1.md
@@ -1,0 +1,26 @@
+# Milestone Delivery :mailbox:
+
+> ⚡ Only the GitHub account that submitted the application is allowed to submit milestones. 
+> 
+> Don't remove any of the mandatory parts presented in bold letters or as headlines! Lines starting with `>`, such as this one, can be removed.
+
+**The delivery is according to the official [milestone delivery guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/milestone-deliverables-guidelines.md).**  
+
+* **Application Document:** https://github.com/w3f/Grants-Program/blob/master/applications/dotnix-follow-up.md 
+* **Milestone Number:** 1
+
+**Context** (optional)
+With this deliverable, Dotnix provides SELinux (Security-Enhanced Linux) integration for secure Polkadot validator operations on NixOS.
+SELinux is a kernel-level mandatory access control (MAC) system that confines processes and resources to explicitly
+defined policies, providing an additional layer of security beyond traditional discretionary access controls.
+
+**Deliverables**
+
+| Number | Deliverable | Link |
+| ------------- | ------------- | ------------- |
+| 0a. | License | https://github.com/sporyon/dotnix-core/blob/main/LICENSE |
+| 0b. | Documentation | https://github.com/sporyon/dotnix-core/blob/main/README.md |
+| 0c. | Testing and Testing Guide | https://github.com/sporyon/dotnix-core/tree/main/checks | 
+| 0d. | Docker | https://github.com/sporyon/dotnix-core/blob/main/docker.nix |
+| 1. | SELinux | https://github.com/sporyon/dotnix-core/tree/main/pkgs/selinux |
+| 2. | Usability | | 


4z3:
Hi @Lederstrumpf,

thanks for your review.

> Both the polkadot.nix dependency & nixpkgs are very outdated, (both March '25, so polkadot-sdk 2412-3, i.e. two generations of stable releases behind, without many crucial backported fixes), so this isn't viable in production.

That's right. We initially tracked `nixpkgs-unstable` the same way `polkadot.nix` does, but this proved difficult to sustain due to API changes and frequent large package rebuilds. To address this, we opted to pin dependencies—allowing us to focus specifically on SELinux integration. While we could rely solely on `flake.lock` for pinning, defining them explicitly in `flake.nix` makes our intent clearer.

There is currently no automated system in place to reliably bump dependencies. We had originally planned to build one as part of https://github.com/w3f/Grants-Program/pull/2439, but that grant application was superseded by this one, which focuses exclusively on security-related aspects.

That said, we've incorporated [your polkadot.nix upgrade](https://github.com/sporyon/dotnix-core/pull/1)—thanks for that!—along with the [the necessary nixpkgs upgrade](https://github.com/sporyon/dotnix-core/commit/acc937e), which brings us to the latest stable NixOS 25.05 release (dated Sun Sep 7 13:27:14 2025 +0200).

> the SELinux setup only works for the strictly default setup. In particular, p2p, rpc, and prometheus ports are hardcoded. Users of dotnix may have different preferences/constraints, and more crucially, this hardcoding prevents multi-validator setups on the same instance. In addition to the network p2p port here, please review the overall SELinux implementation (and dotnix in general) for multi-validator support.

Implementing native multi-validator support in Dotnix would require replacing the current `dotnix.polkadot-validator` options with `dotnix.polkadot-validator.<name>`. While this may be desirable, it is out of scope for the current milestone. The validator service was designed as a singleton, reflecting the convention followed by most services in nixpkgs—and based on the initial assumption that multiple instances would not be needed on the same system.

Perhaps the most idiomatic approach to running multiple instances of singleton services—while annoying—is to use NixOS containers. While containerized instances could theoretically use the same ports and paths, this isn’t supported by default in NixOS: containers compete for ports, and `systemd-nspawn` constructs `CREDENTIALS_DIRECTORY` using `/run/host/credentials` instead of `/run/credentials`. As a result, both ports and paths must be made configurable, which has been done in https://github.com/sporyon/dotnix-core/commit/81747f1 and https://github.com/sporyon/dotnix-core/commit/21c20de, respectively.

As for SELinux, each instance requires distinct types for domains, objects, and ports to avoid collisions. These have been made configurable in https://github.com/sporyon/dotnix-core/commit/dc04ba5 https://github.com/sporyon/dotnix-core/commit/895863e, and https://github.com/sporyon/dotnix-core/commit/6df61a6, respectively.

While these changes alone aren't sufficient, they lay the necessary groundwork for supporting multiple instances on the same system.

In the context of NixOS containers, SELinux configurations are ineffective—any generated policies must be incorporated into a global policy outside the container¹. When doing so, there's no need to worry about duplicate rules or definitions: SELinux will simply apply the first occurrence it encounters in the policy.  While this behavior is usually a nuisance, in this case it's a welcome effect.

For a native solution, the `polkadot-validator.nix` module would require a significant rewrite to support multiple instances.

Note that service users don't necessarily need to differ between instances, as access to their resources is governed by the assigned SELinux contexts.

> Your original application's [acceptance criteria for M1](https://github.com/w3f/Grants-Program/blob/master/applications/dotnix.md?plain=1#L173) include both linux-x86_64 and linux-aarch64 architecture support, yet linux-x86_64 is [hardcoded](https://github.com/sporyon/dotnix-core/blob/main/flake.nix#L51) all over the place. Please address.

This has been addressed in https://github.com/sporyon/dotnix-core/compare/9fe989c%5E...ba1ecdfa0b2e271ab6ca47c3a628d80561e9b825, and https://github.com/sporyon/dotnix-core/compare/23adacb%5E...2cf2c02e195bb54a4b6d7f6d9187353c6bb8b13d.

—
_¹ This limitation is also the reason the provided Docker image was changed to start a VM rather of running the system directly._

Lederstrumpf:
Hi @4z3

Thanks for your detailed feedback & reasoning; apologies for the very belated response on my part.

> There is currently no automated system in place to reliably bump dependencies. We had originally planned to build one as part of w3f/Grants-Program#2439, but that grant application was superseded by this one, which focuses exclusively on security-related aspects.

Acknowledged regarding the cut down scope. I did not expect an automated system in place as a result, but I'd expect somewhat more frequent updates if you're dogfooding this.

> While this may be desirable, it is out of scope for the current milestone. The validator service was designed as a singleton, reflecting the convention followed by most services in nixpkgs—and based on the initial assumption that multiple instances would not be needed on the same system.

I note your point regarding singleton convention with nix services, but just a reminder that running multiple validators on the same bare-metal instance is common (especially for distinct networks), and your grant application promised to be "modular [...] so it can be used by a variety of Polkadot validators using NixOS" (and both milestone objectives are stated as "Security Improvements to all Nix based Polkadot Validators.").

But enough being pedantic; here are two suggestions:
1. Would it help to only do this (native validator support) within the second milestone only? I feel that the later a refactor to a multi-validator setup is performed, the more cumbersome it will become. But as long as it gets delivered eventually, I don't want to block on the work already done.
2. Would it be materially easier to switch to home-manager user services for operating the validators? On a high level, that would easily resolve the issue of parallel service instantiations (using distinct home-manager users for each spawned validator). There may be dragons with SELinux integration I'm unaware of.

---

Continuing with review of your changes: Thanks for the aarch64 integration and making the SELinux policy more flexible - changes largely look good to me, but I stumbled into some issues when testing:

1. Trying to run the checks both on `x86_64` and `aarch64`, I get errors like the following on nixos:
     ```log
       last 25 log lines:
       > alice: (connecting took 12.89 seconds)
       > alice: (finished: waiting for the VM to finish booting, in 13.09 seconds)
       > alice # [   11.984385] AES CTR mode by8 optimization enabled
       > alice # [   12.291103] Console: switching to colour dummy device 80x25
       > alice # [   12.308399] [drm] Found bochs VGA, ID 0xb0c5.
       > alice # [   12.308401] [drm] Framebuffer size 16384 kB @ 0xfd000000, mmio @ 0xfebd0000.
       > alice # [   11.985682] dbus-daemon[645]: dbus[645]: Unknown username "systemd-timesync" in message bus configuration file
       > alice # [   12.341460] [drm] Found EDID data blob.
       > alice # [   12.006934] systemd[1]: Starting Virtual Console Setup...
       > alice # [   12.353121] [drm] Initialized bochs-drm 1.0.0 for 0000:00:02.0 on minor 0
       > alice # error: undefined variable 'x86_64-linux'
       > alice #        at «string»:1:1:
       > alice #             1| x86_64-linux
       > alice #              | ^
      ```
      Full log: 
[flake-check.log](https://github.com/user-attachments/files/23111169/flake-check.log)

      `aarch64` is same, just `s/x86_64-linux/aarch64-linux`

2. When building on nixos-unstable with aarch64-linux (following your example.nix, albeit omitting the docker & container sections), I get errors patching selinux:
      ```log
       > patching file src/systemctl/systemctl-edit.c
       > applying patch /nix/store/s1h9ijczidxcvwhh02y5dinx6b97byfm-0017-meson.build-do-not-create-systemdstatedir.patch
       > patching file meson.build
       > applying patch /nix/store/j6lp6n92pb5k8dmfcngqxdjh8m4mxpii-0018-meson-Don-t-link-ssh-dropins.patch
       > patching file meson.build
       > applying patch /nix/store/d6gi2r8aw5djhfnd8izxbhnhmp7d9cs6-0019-install-unit_file_exists_full-follow-symlinks.patch
       > patching file src/shared/install.c
       > applying patch /nix/store/m0zpma2441b5j7n015i481jhf7inbc1p-0020-timesyncd-disable-NSCD-when-DNSSEC-validation-is-dis.patch
       > patching file src/timesync/timesyncd.c
       > applying patch /nix/store/3anyd8pcn7kwdiccc3zcz7wnrrmj05vv-selinux-label.patch
       > patching file src/core/selinux-setup.c
       > Hunk #1 FAILED at 69.
       > 1 out of 1 hunk FAILED -- saving rejects to file src/core/selinux-setup.c.rej

       For full logs, run:
         nix log /nix/store/ddp745h6axiy16mw3w80b6ydb286zdgv-systemd-257.8.drv
       error: Cannot build '/nix/store/6bdy6hlbzr3if6py025pa9iwzvzs3hm0-extra-utils.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/v40rm4lq72xd30gn0hjyimk0mkp6w5ay-extra-utils
      ```
      I reckon that's due to recent changes Poettering pushed though (https://github.com/systemd/systemd/commit/83b6ef9b62765b11bc602eae906ff13a5464a638) and isn't a milestone acceptance blocker for me - just noting you'll probably run into it on stable eventually.

3. With respect to the network ports, I believe you're only enforcing that specific domains can bind to a particular socket, and have pretty open policies regarding connection? At least when I was running a dotnix validator, I could still rotate its keys with an rpc call from another non-root user without any policy violation. On a secure validator, I reckon the RPC port should be restricted to root users / validator.

4. The `main` script is, well, to put it diplomatically, a little cumbersome ;)
https://github.com/sporyon/dotnix-core/blob/e26ea7b059fd0682c6228567ccce1197682b3d4e/nixosModules/polkadot-validator.nix#L243-L455.
Any plans to replace this? My main gripe is that it's quite user unfriendly, there's isn't even a `--help` function (must look up in source, and the errors returned by it are by its nature not illuminating. For instance, when trying to launch the validator without network keys set, it just reverts
      ``` bash
      polkadot-validator --restart
      Job for polkadot-validator.service failed because the control process exited with error code.
      See "systemctl status polkadot-validator.service" and "journalctl -xeu polkadot-validator.service" for details
      ```
      with the journal log only providing a slightly more indicative hint of
      ```bash
      ...: polkadot-validator.service: Failed to set up credentials: Protocol error
      ...: polkadot-validator.service: Failed at step CREDENTIALS spawning /nix/store/qr7acnwwlbdlzj0glvl36mjvhq48pdf0-polkadot-2506-2/bin/polkadot: Protocol error
      ```
      that the issue is the following line of the service config: https://github.com/sporyon/dotnix-core/blob/main/nixosModules/polkadot-validator.nix#L959-L961

Lederstrumpf (changes requested):
see above comment

4z3:
Hi @Lederstrumpf & thanks for your review!

> I note your point regarding singleton convention with nix services, but just a reminder that running multiple validators on the same bare-metal instance is common [...] Would it help to only do this (native validator support) within the second milestone only? I feel that the later a refactor to a multi-validator setup is performed, the more cumbersome it will become. But as long as it gets delivered eventually, I don't want to block on the work already done.

I agree that adding multi-validator support earlier is better. And there's really no reason to postpone it past this grant except for not anticipating it in the first place. Let's deliver it as part of the second milestone.

> Would it be materially easier to switch to home-manager user services for operating the validators? On a high level, that would easily resolve the issue of parallel service instantiations (using distinct home-manager users for each spawned validator). There may be dragons with SELinux integration I'm unaware of.

home-manager user services are basically systemd user services.  While nice for stuff like desktop environments and other stuff used by interactive users, I'm not sure how well they're suited for system level services, which polkadot very much seems to be. From the top of my head I see at least following annoyances:

- NixOS doesn't provide an automatic restart mechanism when services have changed like it does for system services.
- I'm not aware whether user services start automatically without a user session.

And especially when also defining distinct users just to launch these user services it feels more like a hack than a clean solution.

Instead, I'd go for the previously hinted approach of using `dotnix.polkadot-validator.*`, which is a widely used pattern in nixpkgs.  See e.g.

- `services.anubis.instances.*`
- `services.authelia.instances.*`
- `services.redis.servers.*`
- etc.

> Trying to run the checks both on `x86_64` and `aarch64`, I get errors like the following on nixos: [...]

Not sure how this one could slip through.   It's wrong usage of `nix-build`'s `--arg` parameter, which treats the subsequent argument as Nix code.  The reason this error (let's call it a typo) got introduced is that `jq`'s `--arg` parameter works exactly the opposite way and treats the subsequent argument as string... :)

Either way, the error has been fixed in https://github.com/sporyon/dotnix-core/commit/78a40a938d74b834908f5e39aed0553e0f6f2fe0.

> When building [systemd] on nixos-unstable [...]

Failures like this are to be expected as long as we don't patch systemd upstream—which we might attempt once the SELinux find widespread use in nixpkgs. That said, the fix in this case would basically be a `s/setcon_raw/sym_setcon_raw/` in the patch, which we will be doing when porting to 25.11

> I believe you're only enforcing that specific domains can bind to a particular socket, and have pretty open policies regarding connection? At least when I was running a dotnix validator, I could still rotate its keys with an rpc call from another non-root user without any policy violation.

The policy is pretty strict and grants access to the RPC port only to the `sysadm_systemd_t` domain via this:

    (allow sysadm_systemd_t ${cfg.selinux.rpcPortType} (tcp_socket (name_connect)))

Trying to connect to the RPC port using a regular user produces:

    Oct 27 15:54:57 nixos audit[1304]: AVC avc:  denied  { name_connect } for  pid=1304 comm="curl" dest=9944 scontext=user_u:user_r:user_systemd_t tcontext=system_u:object_r:polkadot_rpc_port_t tclass=tcp_socket permissive=1

Note that using `sudo` doesn't change the caller's context, i.e.

    [root@nixos:~]# id
    uid=0(root) gid=0(root) groups=0(root) context=root:sysadm_r:sysadm_systemd_t

    [root@nixos:~]# sudo -u derp -i id
    uid=1000(derp) gid=993(derp) groups=993(derp) context=root:sysadm_r:sysadm_systemd_t

and neither does `su`:

    [root@nixos:~]# su - derp

    [derp@nixos:~]$ id
    uid=1000(derp) gid=993(derp) groups=993(derp) context=root:sysadm_r:sysadm_systemd_t

So to test the accessibility of the port by another user, the other user needs to log in, which could be accomplished using e.g.

```nix
{
  services.getty.autologinUser = lib.mkForce "derp";
  users.groups.derp = {};
  users.users.derp.group = "derp";
  users.users.derp.isNormalUser = true;
  users.users.derp.password = "derp";
}
```

> The main script is, well, to put it diplomatically, a little cumbersome ;) [...] Any plans to replace this?

Yes, it is cumbersome and there are plans to improve it.  As part of the next milestone of this grant we'll be adding bash completion, and I'll happily add a `--help` option alongside, together with fixes of any user unfriendly messages encountered like the restart without network key you've mentioned.

Lederstrumpf:
Hi @4z3,

Thanks for your detailed reply.

> I agree that adding multi-validator support earlier is better. And there's really no reason to postpone it past this grant except for not anticipating it in the first place. Let's deliver it as part of the second milestone.

> Yes, it is cumbersome and there are plans to improve it. As part of the next milestone of this grant we'll be adding bash completion, and I'll happily add a --help option alongside, together with fixes of any user unfriendly messages encountered like the restart without network key you've mentioned.

Super, great that we're aligned on both of these. Then yes: if multi-validator support and bash-completion/`--help` delivered in the second milestone, all good.
Wrt home-manager services: I agree it's clunkier & less reliable than system services. So please proceed as you suggested with `dotnix.polkadot-validator.*` style.

> Not sure how this one could slip through. It's wrong usage of nix-build's --arg parameter, which treats the subsequent argument as Nix code. The reason this error (let's call it a typo) got introduced is that jq's --arg parameter works exactly the opposite way and treats the subsequent argument as string... :)

> Either way, the error has been fixed in https://github.com/sporyon/dotnix-core/commit/78a40a938d74b834908f5e39aed0553e0f6f2fe0.

Thanks!

> The policy is pretty strict and grants access to the RPC port only to the sysadm_systemd_t domain via this:

> (allow sysadm_systemd_t ${cfg.selinux.rpcPortType} (tcp_socket (name_connect)))
> Trying to connect to the RPC port using a regular user produces:

> Oct 27 15:54:57 nixos audit[1304]: AVC avc:  denied  { name_connect } for  pid=1304 comm="curl" dest=9944 scontext=user_u:user_r:user_systemd_t tcontext=system_u:object_r:polkadot_rpc_port_t tclass=tcp_socket permissive=1

Not the case for me, logged in as a normal user:

❯ id
uid=1009(lederstrumpf) gid=100(users) groups=100(users) context=system_u:system_r:init_t
❯ curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "author_rotateKeys", "params":[]}' 127.0.0.1:9944
{"jsonrpc":"2.0","id":1,"result":"..."}%

with no corresponding avc log (with your example config).

4z3:
Thanks for catching that!

> context=system_u:system_r:init_t

This was most likely caused by logging in via sshd that didn't set SELinux  context, yet. Previously, PAM was only configured to set the context for `login`. This has been fixed in https://github.com/sporyon/dotnix-core/commit/3b7109add350469fb1e7a0b4afead04228cb8678.

